PR #10, opened 2017-08-24
Fix docs: Clarify use of @Prop as attribute

dmorosinotto:
I noticed that if you have a "composed" `propName` and want to set is inside HTML (as an attribute) you need to hyphenated it as `prop-name` or it'll not be set, but inside JSX you can use directly the `propName` if you want.
The `prop-name` syntax works in both cases, so maybe it'll be preferred.


adamdbradley:
Thanks for submitting the PR, but it's something we realized is a bit off internally so we're fixing up props/attributes, then we'll improve these docs. Thanks!

PR #11, opened 2017-08-24
Fix docs @Listen sample to use event.detail

dmorosinotto:
I tested the sample and noticed that the event.`data` doesn't exists! 
I found that to correctly access to the payload emitted in the CustomEvent you need to read `event.detail`


dmorosinotto:
Thanks @mhartington for merging it, I was waiting for some comments about my PRs from @jgw96 from some days... but I think that you are all busy with this new product release and future integration with Ionic!
I'm really looking to see/use it!

PR #43, opened 2017-09-10
Fix inline js not marked as expression in jsx

marvinhagemeister:
As reported in https://github.com/ionic-team/stencil/issues/143 the map statement is laking the braces to mark them as an expression statement in jsx.


mhartington:
Thanks for the PR!

PR #64, opened 2017-10-02
chore(contributing): fix typo and punctuation

peterennis:


peterennis:
Text from old OR still hanging around. I will make a new one.

PR #68, opened 2017-10-04
Fixes loops example

noahlaux:
This PR fixes the templating loops example, as the original doesn't return the inner body of the map function.


jgw96:
Huge thanks for the PR! 🎉 🎉

Issue #71, opened 2017-10-06
Partially wrong linking of Next button or missing buttons


dkundel:
Hi everyone,

I was reading through the docs and I realized that the `Next` button on the [Component Life Cycle page](https://stenciljs.com/docs/component-lifecycle) links to the [Stencil Config page](https://stenciljs.com/docs/stencil-config) rather than [Forms](https://stenciljs.com/docs/forms). 

Additionally are the buttons `Next` and `Back` completely missing on the pages [Forms](https://stenciljs.com/docs/forms), [Updating Arrays and Objects](https://stenciljs.com/docs/handling-arrays) and [Unit Testing](https://stenciljs.com/docs/testing).

I'm happy to submit a PR to fix this.

Cheers,
Dominik

jgw96:
Thanks for the PR, just merged it! 🍻

dkundel:
No problem and thanks for the quick merge :)

PR #72, opened 2017-10-06
Fix links of next/back buttons on several doc pages

dkundel:
Fixes the wrong links described in #71.


jgw96:
Huge thanks for the PR!

PR #85, opened 2017-11-07
docs() update state management examples for clarity

camwiegert:
This updates the code examples in "Managing Component State" to use spread instead of `concat()`. In my opinion, spread makes it clearer that you're assembling a new array. Plus, you don't have to rely on the knowledge that `concat()` doesn't mutate. Definitely open to counter-points.


jgw96:
Thanks for the PR! 🎉

PR #89, opened 2017-11-08
Hide contents inside header on scroll

mbj36:
Content were visibile inside header on scroll
screen shot 2017-11-09 at 2 22 43 am


jgw96:
Hello! Thanks for another PR. I just discussed this with our design person and at the moment, this is a design decision that we are sticking with for now.

PR #91, opened 2017-11-13
docs(router): correct usage of template literals

bjornarg:
The template literal usage in the documentation is not surrounded by `{}` when used in the JSX which yields a compilation error.


jgw96:
Thanks for the PR!

PR #117, opened 2018-01-25
Fix watch docs

adamhjk:
The arguments to functions using `@Watch` is backwards. It was `oldValue, newValue`; it should be `newValue, oldValue`.

Riding along is a second commit that adds the `tags` file to `.gitignore`, for those of us still using ctags. 

Thanks!


adamdbradley:
thanks!!

PR #125, opened 2018-02-08
docs(distribution): fix paths in package.json

dkundel:
Right now the paths are wrong (outdated?) and this wouldn't build.


jgw96:
Thanks for the PR!

Issue #126, opened 2018-02-09
Add Prop mutable docs


brandyscarney:
`@Prop({ mutable: true }) value: string;`

marinho:
I'm working on this as a first contribution.

jgw96:
Closing this as it was added with the above PR. Thanks @marinho for the PR!

PR #146, opened 2018-03-08
docs(prop): Add Prop mutable docs

marinho:
Added documentation about **mutable** option in `@Prop` decorator as requested in https://github.com/ionic-team/stencil-site/issues/126.

Review over English grammar is very welcome.

### DIFF
--- a/src/docs-md/basics/decorators.md
+++ b/src/docs-md/basics/decorators.md
@@ -8,7 +8,7 @@ Each Stencil Component must be decorated with an `@Component()` decorator from t
 
 Use a relative url to the `.scss` file for the styleUrl(s).
 
-```typescript 
+```typescript
 import { Component } from '@stencil/core';
 
 @Component({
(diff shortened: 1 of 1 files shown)


src/docs-md/basics/decorators.md
```diff
@@ -80,19 +80,35 @@ in JSX you set an attribute using camelCase:
...
 console.log(todoListElement.myHttpService); // MyHttpService
 console.log(todoListElement.color); // blue
 ```
 
-It's important to know, that `@Prop` is immutable from inside the component logic. Once a value is set by a user, the component cannot update it internally.
+### Prop value mutability
+
+It's important to know, that `@Prop` is _by default_ immutable from inside the component logic. Once a value is set by a user, the component cannot update it internally.
+
+However, it's possible to explicitly let a `@Prop` allow its value to be modified from inside the component, by declaring it as **mutable**, as in example blow:
```
jgw96: `However, it's possible to explicitly let` should be `However, it's possible to explicitly allow`

marinho: done. I rephrased as `However, it's possible to explicitly allow a @Prop to be mutated...`

marinho:
I updated the code after review. Please re-evaluate.

jgw96:
Thanks for the PR @marinho !

PR #159, opened 2018-04-03
docs(decorators): typo and pluralization

joestrouth1:
Small typo "blow" changed to "below". Added a definite article for clarity/structure.

Modified pronoun so that it agreed with its antecedent in number. Components (plural) specify which props _they_ expect to receive. One component would specify which props _it_ expects to receive.


adamdbradley:
ha, thanks

PR #164, opened 2018-04-08
Fix typo in package name

Serabe:
Didn't create an issue first as the contributing suggest because I think it would be a bit overkill for a small typo. Let me know if it is really needed.


manucorporat:
Thanks @Serabe!
Hope too see you more over here!

Issue #35, opened 2017-09-03
Routing related issue on the site


d2s:
After pressing back button few times at the Stencil website, I ended up to the URL https://stenciljs.com/docs/building-components that looked nothing like the normal site. Apparently route was looping the same home page component again and again.

Screenshot shows what the page looked like in Firefox:
![screenshot from 2017-09-03 13-59-36 stenciljs com - odd routing problem issue](https://user-images.githubusercontent.com/135053/30002370-4096e488-90b0-11e7-9cc8-507d4bfea8dd.png)

Also tried to load the URL with the Chrome browser, and it kept loading indefinitely without the main page content becoming visible.

_Browser version:_
Firefox 55.0.2

jgw96:
Hello @d2s , mind giving this another try? I cant reproduce with the latest version of the site

d2s:
@jgw96 I tried it again (on a different computer).

- Firefox 55.0.3 (64-bit) shows the page URL https://stenciljs.com/docs/building-components as following (screenshot image from the browser).

![screen shot 2017-09-08 at 19 06 39 stencil site bug](https://user-images.githubusercontent.com/135053/30220814-a8b102e8-94c9-11e7-840d-c586bcf323cd.png)

- Both of the latest Chrome and Safari versions keeps loading the https://stenciljs.com/docs/building-components page URL, but do not show any page contents (besides of an expanding scrollbar on the right side of otherwise empty page).

I don't remember what page I was looking before the browser ended up to that URL, so it is kind of difficult to reproduce that part of the issue. What I know is that the mentioned URL causes looping of the content.

jgw96:
Ahh, you're hitting a dead link there. I think you're looking for https://stenciljs.com/docs/my-first-component. I need to get firebase to purge those url's.

janpio:
Same happens with any unkown URL I think: https://stenciljs.com/docs/foobar

PR #172, opened 2018-04-17
Fix wrong link to router docs

EnricoMonese:
Link to __router docs__ points to `/docs/routing` instead of `/docs/router` which results in effects like shown in #35


adamdbradley:
thanks!

PR #177, opened 2018-05-02
Add Ember to framework integration guide

alexlafroscia:
I recently build [`ember-cli-stencil`](https://github.com/alexlafroscia/ember-cli-stencil) which makes it really, really easy to use Stencil components in an Ember app. With the addon installed, and at least one Stencil collection in your dependencies, it will handle all the additional work around discovering where the files are, importing them correctly and copying the lazy-loaded files into the right location in the `dist`. It also supports some "glue code" for better compatibility with older Ember versions ([see wiki for details](https://github.com/alexlafroscia/ember-cli-stencil/wiki/Ember-component-wrappers)).

I decided to just show the absolute minimal guide in the guide here and to instead redirect people to the addon README for additional details. If there's more information you want me to include on your docs site, I'm happy to elaborate more (although there really isn't much more to say besides "install the addon and it does everything else for you!")

### DIFF
--- a/src/docs-md/framework-integration/ember.md
+++ b/src/docs-md/framework-integration/ember.md
@@ -1,0 +1,15 @@
+# Ember
+
+Working with Stencil components in Ember is really easy thanks to the `ember-cli-stencil` addon. It handles:
+
+- Importing the required files into your `vendor.js`
+- Copying the component definitions into your `assets` directory
+- Optionally generating a wrapper component for improved compatibility with older Ember versions
+
+Start off by installing the Ember addon
+
+```bash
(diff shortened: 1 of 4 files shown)


src/docs-md/framework-integration/ember.md
```diff
@@ -0,0 +1,15 @@
+# Ember
```
alexlafroscia: It doesn't seem like these files are being used, but I figured it didn't hurt to add the file here as well as adding to the other page.

jgw96:
Thanks for the PR, we really appreciate it!

PR #184, opened 2018-05-23
docs: avoid "simply"

apetro:
"simply" can discourage a reader who may not yet find the same things to be simple that the documentation authors found to be simple. Therefore avoid using the word "simply" in documentation. Also, this makes the documentation slightly more brief with no loss of meaning.

That "simply" can be off-putting is definitely not my original insight; I regret that I cannot here attribute the source of this advice as I don't recall where I first learned this.


adamdbradley:
Absolutely agree, thanks for pointing that out. Merged!

PR #199, opened 2018-07-18
docs(distribution): correction in import statement for importing published component

LalitKushwah:
import my-component isn't a proper statement. It emphasizes that we are importing any component instead of published project/lib.


adamdbradley:
thanks!

PR #204, opened 2018-08-02
docs(getting-started): fix init command

bitjson:


adamdbradley:
great catch, thanks!!!

PR #210, opened 2018-09-18
Fix import sass plugin

vogloblinsky:


adamdbradley:
Sorry for the delay, but we've got these updates in now. Thanks

PR #217, opened 2018-10-01
sometimes there is nothing to escape which causes unwanted escapes later

Dohxis:
Solves problems like this:
escape problem

Not sure if that's the best solution though, would like to get feedback on this one.


adamdbradley:
thanks

PR #220, opened 2018-10-05
docs: improve event documentation

mattcosta7:
Adds description of the options object accepted by the Event decorator, and specifies that usage in jsx applies to stencil compiled code

fix #218


adamdbradley:
thanks!

PR #232, opened 2018-10-20
docs(@Listen): add key constants info

### LINKED ISSUE #231: Add key constants to Listen info
The constants available for keydown are no longer in the docs, but they are helpful to reference

mattcosta7:
fix #231


manucorporat:
This is mostly a deprecated API, that might be removed at some point in the future, we prefer to do not document it

PR #242, opened 2018-11-02
docs(jsx): add info about using keys in loops

simonhaenisch:
As discussed with @adamdbradley in https://github.com/ionic-team/stencil/issues/1196.
The info is mostly from https://reactjs.org/docs/lists-and-keys.html#keys.


adamdbradley:
thanks!

PR #245, opened 2018-11-03
Automatically deploy commits using Zeit's Now service

simonhaenisch:
This adds a `Dockerfile` and `now.json` to auto-deploy using [Now](https://zeit.co/now) on every commit, i. e. every pull request. See https://zeit.co/github for more info, and https://github.com/simonhaenisch/stencil-global-state-issue/pull/1 as an example PR.

Only thing required by you is to [sign up](https://zeit.co/signup) (you can use Github login) and enable the Github integration for this repository. You can also optionally set up a team (free for OSS). Static deployments are [unlimited and free](https://zeit.co/pricing), however there are monthly storage and bandwidth limits.

We can try this out for a while and see how it goes? I've found it quite useful on other's and personal projects.

---

To test the `Dockerfile` and `now.json`, I kicked of a deploy from my machine locally:

Screen recording of deploying the Stencil site using Now

Everything works, here is the deployment: https://stencil-site-coeggrhaln.now.sh.

---

Just note for future reference: It's possible to set up the domain so that commits to the master branch get auto-aliased (i. e. when merging PRs). Here is the info on how to set up the domain: https://zeit.co/docs/getting-started/assign-a-domain-name (either adding a CNAME record or switching to the Zeit DNS), and you would need to add `"alias": ["stenciljs.com"]` to `now.json`.


simonhaenisch:
I simplified this config to work with Now v2 using `@now/static-build`, i. e., it outputs the site as a static deployment, which is free with the [new pricing](https://zeit.co/pricing).

It basically creates a unique deployment for every PR (and every new commit), and provides its URL as the "details" link in the checks, so you basically have a working (public) preview of every PR. It doesn't do anything when merging a PR into master, and it doesn't interfere with the Firebase deploys.

If you want, you can give it a try. All you need to do is connect the Now Github integration and enable it for this repository.

If you don't want this and only want the auto-deploys from master to Firebase, please just close this PR, so I'll get it off my mind :nerd_face:

adamdbradley:
Thanks for the PR, but we've since moved to auto deploying to firebase.

PR #252, opened 2018-11-15
Fix properties table for Component decorator

FdezRomero:
Hello! 👋

I was browsing the docs and noticed that the `@Component` decorator properties were being displayed as inline text in TypeScript syntax, so I took the liberty to reformat them as a table (which looking at the code seemed that was the intention).

I'm not sure how you'd like to fill the description column, so I skipped it.

Thanks!
Rodrigo

### DIFF
--- a/src/docs/components/decorators.md
+++ b/src/docs/components/decorators.md
@@ -26,16 +26,17 @@ export class TodoList {
 }
 ```
 
-| name | type | description
-tag: string;
-styleUrl?: string;
-styleUrls?: string[] | d.ModeStyles;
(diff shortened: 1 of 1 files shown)


adamdbradley:
Looks like some of the Watch decorator description was removed. Is there a reason to remove that?

FdezRomero:
Sorry about that! I'm used Github's "quick file edit" and somehow I deleted those lines by mistake 😅
I just added them back, thanks for catching it.

adamdbradley:
awesome thanks!

PR #255, opened 2018-11-16
Use auto binding arrow function syntax

danawoodman:
Just a suggestion, but encouraging people to use the cleaner auto-binding arrow function syntax instead results in cleaner code (IMO). Feel free to accept/reject as you see fit!

Also made it consistently use `e` for the event param

### DIFF
--- a/src/docs/components/forms.md
+++ b/src/docs/components/forms.md
@@ -21,22 +21,22 @@ export class MyName {
 
   @State() value: string;
 
-  handleSubmit(e) {
+  handleSubmit = (e) => {
     e.preventDefault()
     console.log(this.value);
     // send data to our backend
   }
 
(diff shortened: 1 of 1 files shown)


manucorporat:
I would rather explain this better in a "Best practices" guide, otherwise people might misunderstand they can use the method directly without `bind()` and that could break their apps.

danawoodman:
True, but previous examples use this syntax so if so those should also be changed. I did this in part to keep the examples more consistent

adamdbradley:
Sorry we were unable to merge this, but looks like too many conflicts. Thanks

danawoodman:
Lol, yeah that’s because you guys waited 5 months to merge it! 😂

PR #257, opened 2018-11-19
Adds disabling service worker generation

DavidFrahm:
Helping users know how to use Stencil Config to disable the automatic generation of a service worker file.

Please let me know if the way I worded this seems helpful and consistent with other docs.

### DIFF
--- a/src/docs/tooling/config.md
+++ b/src/docs/tooling/config.md
@@ -349,3 +349,12 @@ export interface EmulateViewport {
 
 }
 ```
+
+## Disable service worker
+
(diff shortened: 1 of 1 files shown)


DavidFrahm:
Just realized I created the issue for this in the `stencil` repo:

https://github.com/ionic-team/stencil/issues/1229

manucorporat:
I am not sure this is the right spot to add this. There is a service-workers file: https://github.com/ionic-team/stencil-site/blob/master/src/docs/tooling/service-workers.md

manucorporat:
I added your changes in a different commit!

DavidFrahm:
@manucorporat Thanks, but where? I don't see them in the docs yet.

PR #272, opened 2018-12-04
docs: example using components with ViewChild in Angular

peterpeterparker:
Following our discussion on the Slack channel about how to consume Web Components build with Stencil in an Angular project, we had the idea to add an example about this in the documentation

Feel free to improve it


peterpeterparker:
thx @manucorporat

PR #279, opened 2018-12-05
docs(unit-testing): remove async from example

bassettsj:
- remove async from code example that doesn't use any await keywords


bassettsj:
So fast! Thanks for the quick review.

PR #295, opened 2019-01-01
docs(config): typo

flawyte:
Btw, it's unclear what the relation is between this file and using a css pre-processor. This is why I left the "can be", but imo should use "is" if the file is **required** for CSS pre-processing via Stencil.


adamdbradley:
thanks

PR #299, opened 2019-01-07
Stop using test-components name

JackuB:
1. `test-components` is actually an unrelated npm package https://www.npmjs.com/package/test-components
1. it sucks when you copy paste a demo code to jsbin/local env and it spews an error

Issues with this PR:

- had to use the `/esm` path to expose `defineCustomElements` https://github.com/ionic-team/ionic/issues/14925
- fixing `@latest` in unpkg url is bad, but v4 is RC only, so can't do `@4`

### DIFF
--- a/src/docs/framework-integration/angular.md
+++ b/src/docs/framework-integration/angular.md
@@ -50,7 +50,7 @@ import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { AppModule } from './app/app.module';
 import { environment } from './environments/environment';
 
-import { defineCustomElements } from 'test-components/dist/loader';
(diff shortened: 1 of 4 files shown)


src/docs/framework-integration/angular.md
```diff
@@ -50,7 +50,7 @@ import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { AppModule } from './app/app.module';
 import { environment } from './environments/environment';
 
-import { defineCustomElements } from 'test-components/dist/loader';
```
manucorporat: the `loader` part is the most important part, esm should not be used.

PR #301, opened 2019-01-12
docs: my-first-component without CSS

JackuB:
Either mention you need to create it or drop it from the example, otherwise you are greeted with:

```
[ ERROR ]  Unable to load style src/components/my-first-component.css from src/components/my-first-component.tsx
```


adamdbradley:
thanks

PR #316, opened 2019-02-14
Proposal: add note for Stencil ready custom event

dangodev:
I wanted to propose an addition to the docs to solve a problem I encountered. I wanted a way to know when Stencil was ready, and Anthony Johnston from the [Slack channel](https://stencil-worldwide.slack.com/archives/C79EANFL7/p1549876428005900?thread_ts=1549659988.094200&cid=C79EANFL7) shared this snippet with me.

I wanted to add a note, as this was the best method I could find to wait for Stencil to render. Feel free to propose adding this to a different section, or any edits. I just wanted it to be in the docs _somewhere_!

### DIFF
--- a/src/docs/framework-integration/javascript.md
+++ b/src/docs/framework-integration/javascript.md
@@ -94,3 +94,20 @@ export class TodoList {
 ```tsx
 <todo-list my-object="{}" my-array="[]"></todo-list>
 ```
+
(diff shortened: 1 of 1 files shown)


adamdbradley:
This is an api still being worked on, and may get removed, so for now we'll probably keep it undocumented. Thanks

PR #324, opened 2019-02-25
docs: Correct fields in `package.json` example

chris-dura:
The `main` and `types` field were not accurate to what is actually generated by the `stencil-component-starter`.

I'm not _absolutely_ sure which is correct, but I know, following the documentation, and publishing to NPM... my `dist` definitely does not have a `dist/myname.js` or a `dist/types/index.d.ts`...


adamdbradley:
thanks!

PR #325, opened 2019-02-28
Working with conditionals and the key-attribute to recreate components

cmaas:
Clarifies a problem I had that lead to somewhat unexpected behavior. I've seen the branches in a conditional as different siblings and not the very same component. The updated docs make clear how to destroy and recreate components if a conditional switches the branch.


adamdbradley:
good stuff, thanks!

PR #335, opened 2019-03-21
Update vue.md

glennlawyer:
Wanted to flesh out the comments on how to use Stencil with Vue.

Thanks to 

https://vuejs.org/v2/api/#ignoredElements

https://medium.com/@ales.genova/stenciljs-in-react-vue-angular-82fded0c738f
https://github.com/alesgenova/stenciljs-in-vue/blob/master/src/main.js

### DIFF
--- a/src/docs/framework-integration/vue.md
+++ b/src/docs/framework-integration/vue.md
@@ -12,42 +12,40 @@ contributors:
 
 # Vue
 
-In order to use the custom element library within the Vue app, the application must be modified to define the custom elements and to inform the Vue compiler which elements to ignore during compilation. This can all be done within the `main.js` file. For example:
+In order to use the custom element library within the Vue app, the application must be modified to define the custom elements and to inform the Vue compiler which elements to ignore during compilation. This can all be done within the `main.js` file. 
 
-### Add the component(s) to the app dependencies
-Typically done with `npm install --save test-components`
-
-
-### Import the components into the 'main.js' file
-by 
+Assuming you’ve run `npm install --save test-components` beforehand, and that `test-component` is the name of our made up Web Components that we have published to npm, you import the components into the 'main.js' file by 
 
 - importing the node module
-- telling Vue to ignore the custom element tags
-- binding their code to the window object
-
-See `// HERE` comments:
+- telling Vue to ignore the custom element tags (see `https://vuejs.org/v2/api/#ignoredElements`)
+- binding the Stenciljs component code to the window object
 
 ```tsx
 import Vue from 'vue';
 import App from './App.vue';
-import { defineCustomElements } from 'test-components/dist/loader'; // HERE
+
+import { defineCustomElements } from 'test-components/dist/loader'; 
 
 Vue.config.productionTip = false;
-Vue.config.ignoredElements = [/test-\w*/]; // HERE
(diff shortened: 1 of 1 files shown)


src/docs/framework-integration/vue.md
```diff
@@ -14,19 +14,43 @@ contributors:
...
+
+### Import the components into the 'main.js' file
+by 
+
+- importing the node module
+- telling Vue to ignore the custom element tags
+- binding their code to the window object
+
+See `// HERE` comments:
+
```
dangodev: Just a neutral bystander reviewing here: if I were reading these docs I’d find the incomplete sentences jarring (especially if you’re trying to continue a sentence started in a heading that tries to continue into body text `… into the 'main.js' file … by`).

Further, the previous paragraph ends with “For example:”, and a new heading begins. Seems like the document flow is disrupted.

Instead, I think this entire section could be removed, and the previous sentence could be edited:

```diff
-In order to use the custom element library within the Vue app, the application must be modified to define the custom elements and to inform the Vue compiler which elements to ignore during compilation. This can all be done within the `main.js` file. For example:
+In order to use the custom element library within the Vue app, the application must be modified to define the custom elements and to inform the Vue compiler which elements to ignore during compilation. This can all be done within the `main.js` file. Assuming you’ve run `npm install --save test-components` beforehand, take the following  example:
```

Brevity is clarity!

src/docs/framework-integration/vue.md
```diff
@@ -14,19 +14,43 @@ contributors:
...
 import App from './App.vue';
-import { defineCustomElements } from 'test-components/dist/loader';
+import { defineCustomElements } from 'test-components/dist/loader'; // HERE
 
 Vue.config.productionTip = false;
-Vue.config.ignoredElements = [/test-\w*/];
+Vue.config.ignoredElements = [/test-\w*/]; // HERE
 
-defineCustomElements(window);
+defineCustomElements(window); // HERE
```
dangodev: Personally I find these comments more distracting than helpful. There are only a few lines here; not much to explain.

glennlawyer: Glad for comments from a neutral bystander-- what makes sense to me does not always make sense to everyone.

I like the "HERE" tags because it highlights what gets added to the Vue file. It is not otherwise obvious where the file is modified. 

But I could instead put it in the text?

src/docs/framework-integration/vue.md
glennlawyer: Agreed and thank you. 

I don't know how to get back into my suggested modifications to edit them.

src/docs/framework-integration/vue.md
dangodev: Ah I see. Maybe more explanatory comments would help, like this one from the React integration:
```js
// test-component is the name of our made up Web Component that we have
// published to npm:
import { defineCustomElements } from 'test-components/dist/loader';
```

glennlawyer: Another improvement over how I had put it. Thank you again.

glennlawyer:
@dangodev just pushed a new version which incorporates your suggestions, and closed the conversations. 

Really  appreciate your suggestions, thanks again.

adamdbradley:
looks good thanks

PR #336, opened 2019-03-21
Add example passing external values to page.$eval

dangodev:
This was a head-scratcher for me for weeks! Until someone in the Stencil Slack channel kindly pointed me in the right direction. I had read this page so many times looking for this answer, and found others in the Slack group asking the same.

Adding it here, because at least on our team this was a common complaint against the E2E tests. Some of our components have pretty deep-nested data structures, which means the tests are almost worthless without good mock data.

Open to feedback!


adamdbradley:
Looks good thanks

PR #344, opened 2019-04-12
Make it clearer that the Watch decorator don't fire during the initial component load.

rosdi:
Clarifies that this decorator won't fire during the initial load.


adamdbradley:
thanks!

PR #356, opened 2019-04-23
docs: adjust initial project setup

Jefiozie:
I was looking at Stencil and I found out that the docs missed something that will explain how to start with a new project. This issue is also reported [here](https://github.com/ionic-team/stencil/issues/1397)

If I need to make changes our there needs some more explaining let me know.

btw, a big thanks for this and making it OSS 🙏🏻

### DIFF
--- a/src/docs/introduction/getting-started.md
+++ b/src/docs/introduction/getting-started.md
@@ -14,6 +14,14 @@ Stencil requires a recent LTS version of [NodeJS](https://nodejs.org/) and npm. 
 
 > Note that you will need to use npm 6 or higher.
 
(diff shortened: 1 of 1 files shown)


simonhaenisch:
You don't need to install Stencil globally in order to use the `npm init` command. `npm init stencil` will work even when Stencil is not installed, and create a project folder with a local installation of Stencil. The `npm init` command calls a package called [`create-stencil`](http://npmjs.com/package/create-stencil), you can have a look at the docs of the `init` command here: https://docs.npmjs.com/cli/init#description.

Jefiozie:
Thank you, TIL I never knew this.

PR #366, opened 2019-05-02
Update getting-started.md

SiddharthaKalyanasundaram:
When I tried with npm init stencil app, then only it got worked.


simonhaenisch:
I used `npm init stencil` today and didn't need to add `app` at the end...

adamdbradley:
This may have been a bug but seems to be fixed now. For our getting started we recommend just `npm init stencil` and let the user choose from the interactive output. Thanks

PR #375, opened 2019-05-15
Add port property to devServer-Config

feerglas:
I guess lot of people want to know how they can config a specific port for the devServer...


adamdbradley:
thanks

PR #379, opened 2019-05-16
docs(e2e-testing): Update e2e testing to inform about CI/CD caveats

Nhiokh:
I ran accross some problems when trying to setup my e2e tests on Gitlab CI. Turns out I didn't see that stencil's config offers a way to pass args to the Chrome Headless before running those tests. Therefore, I was wondering how to improve the doc on this particular matter. This is my attempt to do so.


adamdbradley:
thanks

PR #382, opened 2019-05-20
Update docs-auto-generation.md

jordanranz:
- Update outputTarget for docs auto generation based off of [this change](https://github.com/ionic-team/stencil/blob/master/BREAKING_CHANGES.md#deprecations)


adamdbradley:
thanks

PR #388, opened 2019-05-27
Add testing/mocking.md

simonhaenisch:
Adds a new section to "Testing" called "Mocking".

There were also some automatic changes to `components.d.ts` which I wasn't 100% sure about, so I put them in a separate commit (to make it easier to undo, just in case).

Mocking - Stencil


adamdbradley:
awesome thanks!

PR #389, opened 2019-05-30
Add async keyword to example in @Method docs

cthos:
The first example _might_ return a promise, but I figured the first example in the page should be explicit, since there's no method body showing a promise being returned and whatnot.


adamdbradley:
thanks

PR #403, opened 2019-06-04
Fix invalid outputTarget config in docs

jeanbenitez:
The type is defined as object key instead using 'type' property, maybe is a bad copy from previous stencil version

cc @adamdbradley


adamdbradley:
thanks

PR #406, opened 2019-06-05
docs(host-element): explicit about host component

Matthew-Smith:
Modified the `<Host>` as `<Fragment>` section to be explicit about what the component tag name is.
Modify the @Element decorator section to not reference `hostData()`.


adamdbradley:
Thanks for the much needed content updates!

PR #408, opened 2019-06-06
Basic search functionality

seanwuapps:
Add search input field on the `site-menu` component: as user types search terms, the menu items get updated. 
Search is based on the `text` field in a `SiteStructureItem` object


adamdbradley:
Thank you for your PR, but we opted to use algolia instead.

Issue #409, opened 2019-06-06
Wierd double border on contributor list


Carlos-T:
![image](https://user-images.githubusercontent.com/12251313/59038775-e1a7d500-8873-11e9-80af-8616778bc9a3.png)
Both image and image wrapper are setting a border. Although is the wrapper that should set the border, It's quite easier to remove the wrapper's one and left the image's one.

Carlos-T:
Just for learning... Removing the border causes the images to be more close to each other. Also keeping border-radius 100% makes the wrapper oval-shaped. I know this is not critic at all, but I want to learn a bit.

Thanks in advance.

PR #410, opened 2019-06-06
Removed double border

Carlos-T:
Solves  #409 but as said there, maybe is not the perfect solution.


manucorporat:
Thanks Carlos! but we already fixed the double border :)

PR #424, opened 2019-06-08
type(docs): update h1 to match title & menu

jbmoelker:
When following links to the page about "styling" or "styling components" you end up on a page with an h1 "Shadow DOM". This is confusing. The page is about "Styling (Components)" as the title and description in this markdown file also state. The page has two main topics "Shadow DOM" and "CSS Variables", so those should be the h2 level headings. The other sections on Shadow DOM are therefore changed to h3 level headings.


adamdbradley:
good call, thanks

PR #428, opened 2019-06-11
Update mocking.md

zooduck:
When using newSpecPage, you must either set the element on the html property of the options, or use page.setContent() instead.


adamdbradley:
thanks

PR #430, opened 2019-06-12
docs(): new import folder for stencil components

mburger81:
with new stencil one we have to import stenciljs components from test-components/loader, so without the dist folder


adamdbradley:
thanks

PR #431, opened 2019-06-12
Corrected @Prop 'reflect' (false) example

mattezell:
Updated 'reflect' @Prop usage example so that it demonstrates property assignment vs attribute display when reflect is set to false.


adamdbradley:
thanks

PR #434, opened 2019-06-13
Methods should be async

Matthew-Smith:
In accordance with https://stenciljs.com/docs/methods#public-methods-must-be-async


adamdbradley:
good catch, thanks!

PR #436, opened 2019-06-14
docs(): add applyPolyfills to React and Vue integrations.

jthoms1:


Javismomo:
Hi @jthoms1 , could you add this changes to the angular framework integration too, issue is there too and I fixed it appplying these changes.

Thanks!

jthoms1:
Ya I can add there.  The assumption was that Angular already had these polyfills loaded but I guess not.  @manucorporat any thoughts on this.  I know that we do not do this for Ionic right?

PR #439, opened 2019-06-15
List reserved prop names

bitflower:
I ran into this yesterday. Maybe worth adding a compiler warning anyway but this might be a start.


adamdbradley:
Closing for now, the list is changing still

PR #440, opened 2019-06-15
type(docs): fix typo in Host as Fragment

jbmoelker:
It said "`<Host>` can only be used when ...". But clearly it can be used in multiple situations such as the one in the section above with no children.


adamdbradley:
thanks

PR #441, opened 2019-06-16
docs(host): Add some styling notes about the host element

Matthew-Smith:
I encountered some issues where my css was not being applied properly to the `<Host>` Element when I was using Shadow DOM. This adds a little bit of description to the `<Host>` Element page.


adamdbradley:
thanks

Issue #442, opened 2019-06-16
Table of Contents not parsing escaped characters


Matthew-Smith:
<img width="201" alt="Screen Shot 2019-06-16 at 3 08 57 PM" src="https://user-images.githubusercontent.com/767853/59568489-3d7b1680-9049-11e9-84fd-abdffd0f3b3b.png">
<img width="172" alt="Screen Shot 2019-06-16 at 3 08 47 PM" src="https://user-images.githubusercontent.com/767853/59568490-3d7b1680-9049-11e9-8f95-bfcd3e5e99ab.png">

PR #443, opened 2019-06-16
bug(in-page-nav): clean headers before showing

Matthew-Smith:
See https://github.com/ionic-team/stencil-site/issues/442


adamdbradley:
great fix, thanks!

PR #459, opened 2019-06-28
Create local-assets.md

splitinfinities:
Potentially we can add an example of referencing local images in global, and also from css, if needed. 

Open to any and all feedback!


src/docs/guides/local-assets.md
```diff
@@ -0,0 +1,55 @@
...
+  tag: 'stencil-asset',
+  styleUrl: 'stencil-asset.css',
+  assetsDirs: ['assets']
+})
+export class StencilAsset {
+
+  @Prop() src: "sunset.jpg"|"beach.jpg" = "sunset.jpg";
+
+  render() {
+    <img src={`${getAssetPath('.')/assets/${this.asset}`} />
```
simonhaenisch:
```suggestion
    <img src={`${getAssetPath('.')assets/${this.asset}`} />
```

I think `getAssetPath(.)` returns a string ending with `/`?

splitinfinities: That's great to know!

mburger81:
Hello, this is not working for me as expected.

With `npm run start` locally anything is fine, but if I bundle the component and install it trough npm on an ionic 4 project running this on dev with "ionic serve" getAssetPath returns mi the root, so `getAssetPath('./assets/img.svg');` is `/assets/img.svg`
but the image naturally is not there. And honestyla I don't know where the images should be. I know only that on build the components the images are bundled here

```
dist/collection/components/aquapur/assets/aquapur-blue.svg
dist/lan-web-components/assets/aquapur-blue.svg
```

But how can I access them from ionic 4 project?

PR #469, opened 2019-07-17
Update events.md

unipro-wesley:
event.key is part of the KeyboardEvent API, and not "some constants Stencil provides" =)


adamdbradley:
thanks

PR #470, opened 2019-07-24
Update angular.md

rbalet:
1.  Try to improve the import of `CUSTOM_ELEMENTS_SCHEMA`. In fact, only importing it in the module that needs custom elements work fine for evergreen browser, but when you want it to work on Edge/IE11 you **have to** import it into the `AppModule`
2.  Include Edge and IE11 poyfills (Not having this informations may lead to problem)


adamdbradley:
thanks

PR #478, opened 2019-08-20
add info about component generator

simonhaenisch:
image

Will update the starters now to include said script.

**Edit:** just added another paragraph:

> All components will be generated within the `src/components` folder. Within that, a folder will be created with the same name as the component tag name you provided, and within that folder the files will be generated. It is also possible to specify one or multiple sub-folders to generate the component in.
>
> For example, if you specify `pages/page-home` as the component tag name, the files will be generated in `src/components/pages/page-home`.


adamdbradley:
🎉

PR #480, opened 2019-08-21
add info for using a `.vscode/launch.json`

simonhaenisch:
This has been asked and referenced on Slack multiple times, and I already posted this config like four times, so I thought it's time to add it to the docs.

image

The config should probably be tested by someone on Windows, and I can't remember why I set `disableOptimisticBPs`. The tooltip says

> Don't set breakpoints in any file until a sourcemap has been loaded for that file.

and @manucorporat is/was working on getting the sourcemaps working for testing, so we can probably leave it in.

### DIFF
--- a/src/docs/testing/overview.md
+++ b/src/docs/testing/overview.md
@@ -54,3 +54,36 @@ export const config: Config = {
   }
 };
 ```
+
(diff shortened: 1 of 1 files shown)


adamdbradley:
🎉

PR #493, opened 2019-09-10
Update redux.md

stanley85:
Just in case no one will find this either.


adamdbradley:
From the site we'll be recommending stencil/store instead.

PR #501, opened 2019-09-30
fix: add local-assets guide to the menu

simonhaenisch:
Also added myself to the contributors because the examples from the guide were mostly taken from a message I posted on slack once.

BTW someone didn't commit `components.d.ts` after updating to Stencil 1.5.3 so I had quite a few changes there, but I didn't add it to this commit either (because I don't know when this is going to be merged, and wanted to prevent merge conflicts).


adamdbradley:
thanks

PR #507, opened 2019-10-07
add project context

bitflower:
Hey guys,

I was totally missing the project context when looking at the example folder hierarchy. I was mixing up `assets` with a folder `assets` in the root of my Stencil project.

Not sure if this for everyone but it would have helped me.

Matt


adamdbradley:
good call

PR #510, opened 2019-10-08
docs: framework integrations dist/loader

peterpeterparker:
Just in case you would like to display the example of framework integrations with the default `dist/loader` aka

```
import { applyPolyfills, defineCustomElements } from 'test-components/dist/loader';
```

instead of

```
import { applyPolyfills, defineCustomElements } from 'test-components/loader';
```


adamdbradley:
thanks

PR #517, opened 2019-10-22
Update styling.md

matthew-dean:
CSS Custom properties are much more like Less variables than Sass variables, as Less vars (like CSS vars) are declarative and have a value per scope, and Sass variables are imperative and can change values mid-scope (and can also be re-defined as themselves, whereas neither Less nor CSS Variables can be). So Less / CSS have a shared property / variable paradigm, whereas Sass uses something more like a Ruby paradigm, using CSS-like syntax.


adamdbradley:
It's just a simple comparison and think Sass is fine.

Issue #519, opened 2019-11-02
Docs: Server Side Rendering


dgautsch:
The site has documentation for prerendering is missing information on how to integrate the the hydrate app that is generated from the `--prerendering` build option. Having this information will help future developers integrate stencil components into an existing SSR application.

PR #520, opened 2019-11-02
Document How to Use the Hydrate App for SSR

dgautsch:
Addresses #519 

I wasn't sure how far we want to take the documentation. I also know that the `hydrateDocument` takes a large config object and I'm unsure which each option does. I plan on expanding this PR with more information.

Feedback is very welcome!

- [x] Document config object options. Will this be changing? How much of it is worth documenting?
- [x] Document `renderToString`


adamdbradley:
Awesome, love it! Let me do an update quick to give it it's own page

adamdbradley:
https://github.com/ionic-team/stencil-site/commit/5b39bd04558398f3bf53d237763a9b7d3b58b7cb

PR #528, opened 2019-11-16
docs(docs-json): add info for slots and usage

### LINKED ISSUE #527: docs-json output target needs info on slots and usage
Had to do a little digging to figure out that slot name and description are separated by a hyphen. While doing so, I found the very neat usage feature which also needs a bit of info. I've seen some questions on the slack and stackoverflow about that one.

I'll open a PR and add a quick section to the output target.

MattL75:
- Adds sections for slots and usage to docs-json

- Removed the _Documenting_ prefix for CSS variables because adding it for slot and usage seemed pretty redundant and looked kinda weird on the ToC

closes #527


dgautsch:
Nice Addition!

manucorporat:
Great docs!! Thanks a lot!!!!

PR #538, opened 2019-12-04
update forms.md

rahul-rautwar:
Added **selectedReceiverIds** and **avOptions** values so that If anyone copies this code and try to run it then it will work smoothly (earlier it was giving an error for **selectedReceiverIds** in render ).
Also updated **handleSubmit** **method** as in render block we called submit method with parameters


adamdbradley:
thanks

PR #555, opened 2020-01-14
feat: Docs for Extras config

mattdsteele:


mattdsteele:
I was having the hardest time getting the docs to build on my machine, and I don't see a preview URL anywhere, so this is a classic push and pray... probably worth verifying before merging :)

PR #568, opened 2020-02-14
Add example of global script export and mention async

bitflower:


src/docs/config/overview.md
```diff
@@ -54,10 +54,17 @@ enableCache: true
...
+
+```javascript
+export default function() { // or export default async function()
+  initServerConnection();
+}
+```
 
 The global script config option takes a file path as a string.
 
+> The exported function can also be `async`
```
simonhaenisch:
```suggestion
> The exported function can also be `async`.
```

src/docs/config/overview.md
```diff
@@ -54,10 +54,17 @@ enableCache: true
 
 The global script runs once before your library/app loads, so you can do things like setting up a connection to an external service or configuring a library you are using.
 
-The code to be executed should be placed within a default function that is exported by the global script. Ensure all of the code in the global script is wrapped in the function that is exported.
+The code to be executed should be placed within a default function that is exported by the global script. Ensure all of the code in the global script is wrapped in the function that is exported:
+
+```javascript
```
simonhaenisch: Is it still that highlighting only works if this is set to `typescript`?

src/docs/config/overview.md
```diff
@@ -54,10 +54,17 @@ enableCache: true
...
+export default function() { // or export default async function()
```
simonhaenisch: Just a suggestion cause the rest of the docs (e. g. Guides and Framework Integrations) mainly uses arrow functions:

```suggestion
export default () => {
```

(I think it's more concise without the comment)

src/docs/config/overview.md
```diff
@@ -54,10 +54,17 @@ enableCache: true
...
+  initServerConnection();
+}
+```
 
 The global script config option takes a file path as a string.
```
simonhaenisch: Could move this line to be the first sentence of the first paragraph, so that the note about "async" is just below the example.

src/docs/config/overview.md
bitflower: Top :-)

src/docs/config/overview.md
bitflower: Not sure, it looks ok in the preview here.

PR #571, opened 2020-02-17
Fix link

zigomir:


dmartinjs:
I already opened a PR ( #567 ).

The original link is pointing a ts file, 
so I think it should point to https://github.com/ionic-team/stencil/blob/master/src/declarations/stencil-public-docs.ts now

adamdbradley:
thanks for the update, but merged in the previous pr

PR #573, opened 2020-02-22
Update local-assets.md

andrzejtemplin:
Probably it won't work since the 'return' keyword is missing. I have tried it and it doesn't work for me though.


adamdbradley:
good catch, thanks

PR #582, opened 2020-03-11
docs(reactive-data): typos, wording tweaks

rtpHarry:
- some wording tweaks to split sentences up and fix typos
 - added @State() decorators to the code examples to make it clear. I've been caught out a few times and then realised I was updating a plain array/object, and not something Stencil was monitoring.


adamdbradley:
thanks

PR #586, opened 2020-03-14
docs: Async methods must return a Promise.

max-scopp:
This issue has been raised in slack.

This is the latest one, but I've seen it more than once.
https://stencil-worldwide.slack.com/archives/C79EANFL7/p1584186459046200

### DIFF
--- a/src/docs/components/methods.md
+++ b/src/docs/components/methods.md
@@ -51,13 +51,19 @@ Stencil's architecture is async at all levels which allows for many performance 
 
 
 ```tsx
+// VALID: using "async" keyword
+@Method()
+async myMethod() /* returnType: Promise<number> */ {
+  return 42;
+}
+
 // VALID: using Promise.resolve()
 @Method()
-myMethod() {
+myMethod2() {
   return Promise.resolve(42);
 }
 
-// VALID: even if it returns nothing, it needs to be asynchronous
+// VALID: even if it returns nothing, it needs to be async
 @Method()
 async myMethod3() {
   console.log(42);
@@ -65,13 +71,13 @@ async myMethod3() {
 
 // INVALID: must use "async" keyword or return Promise
 @Method()
-notOk() {
+notOk() /* returnType: number */ {
   return 42;
 }
 
-// INVALID: must return Promise
+// ERROR: ts(1064) The return type of an async function or method must be the global Promise<T> type.
 @Method()
-async myMethod() {
+notOk2(): number {
   return 42;
 }
 ```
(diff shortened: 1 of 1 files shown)


max-scopp:
See https://github.com/microsoft/TypeScript/issues/15292

max-scopp:
I agree, but I've updated them to clarify that you can use public methods without the "async" keyword, but it *always* needs an `Promise` returned.

Maybe it should be mentioned that the "async" keyword is still preferred, but that's a general knowledge for EcmaScript IMO.

manucorporat:
```
@Method()
async myMethod4() {
  return 42;
}
```
this is not invalid, this method already returns a promise.

max-scopp:
I've reverted the docs after some clarification on the return types.

But I suggest a small addition related to TypeScript specifically, to indicate that type implicity.

src/docs/components/methods.md
```diff
@@ -69,9 +69,15 @@ async myMethod3() {
...
 
-// INVALID
+// INVALID: must use "async" keyword or return Promise
 @Method()
-notOk() {
+notOk() /* returnType: number */ {
+  return 42;
+}
+
+// ERROR: ts(1064) The return type of an async function or method must be the global Promise<T> type.
```
manucorporat: This example seems off.

I don't think we should include this example because it's not stencil related, it's just how JS and TS. works. I would rather add a note about `async` functions, returning a Promise by spec, so the return type  is a Promise

src/docs/components/methods.md
```diff
@@ -69,9 +69,15 @@ async myMethod3() {
   console.log(42);
 }
 
-// INVALID
+// INVALID: must use "async" keyword or return Promise
```
manucorporat: async already returns a Promise

```suggestion
// INVALID: must use "async" keyword or explicitly return a Promise
```

src/docs/components/methods.md
```diff
@@ -51,9 +51,9 @@ Stencil's architecture is async at all levels which allows for many performance
 
 
 ```tsx
-// VALID: using async
+// VALID: using "async" keyword
 @Method()
-async myMethod() {
+async myMethod() /* returnType: Promise<number> */ {
```
manucorporat: It might cause confusion to mention `returnType` some new developers might thing it's a new thing, i would rather explain in the comment above that a async function/method implicitally returns a Promise, so the return type is `Promise<number>`

src/docs/components/methods.md
manucorporat: ok, looks like it was fixed, but still not sure about this example. It's not valid, not because of stencil, it's not valid, because of TS itself

claviska:
I agree with @manucorporat that we shouldn't show syntactically incorrect examples. Looking at the docs with fresh eyes, they seem pretty clear right now as to how return values in methods should work. This is getting stale, so I'm going to close it out.

PR #587, opened 2020-03-16
docs(styling): typo within / withing

rtpHarry:


rtpHarry:
Oops didn't check first, somebody else has submitted this with #585

PR #590, opened 2020-03-18
docs(style-guide): fix code snippet highlighter

rtpHarry:
Looks like the `...` in the code sample is breaking the syntax highlighter, so I've put comments on them.

image


adamdbradley:
good catch, thanks

PR #592, opened 2020-03-20
feat: window isn't needed anymore to load components with defineCustomElements

peterpeterparker:
As discussed on Slack, `window` isn't needed anymore to load components with `defineCustomElements`.

Moreover, on Gatsby build, no using `window` is a must, otherwise their prerendering build would fail.

That's why, it's better to display `defineCustomElements` without it.


adamdbradley:
thanks!

PR #597, opened 2020-04-15
store.md IE11 compatibility

numerized:
Thanks for amazing way of look at things.
This PR is my proposal to make it easy to comprehend for IE11 compatibility.

Let me know if you like it as is.


numerized:
need more tests I was drunk or ...

PR #598, opened 2020-04-15
add .container class to wrapper

seanwuapps:
Add .container to wrapper div to make the page style look consistent


adamdbradley:
thanks

PR #599, opened 2020-04-16
Doc store ie11 compatibility 

numerized:
I have added a full IE11 example, I tried it it works.


mlynch:
Thanks for the PR! We are moving away from officially supporting IE 11 (though Stencil does currently), so won't be adding an extra example for IE11. Time to move on!

PR #604, opened 2020-04-27
docs(testing): add hint for upstream Jest config

viernullvier:
This change will add a hint explaining that Jest configurations in `package.json` and `jest.config.js` will not be applied. While we're at it, it also fixes two minor punctuation issues.


adamdbradley:
thanks

PR #606, opened 2020-05-06
docs(faq): add more info on browser support

### LINKED ISSUE #499: Browser Support in FAQ
Currently in the FAQ page Explorer 11 is mentioned but no other browser is documented here.

I'd suggest a more detailed browser table with minimum required version.

Chrome 67 >
Firefox 63 >
Safari (partial)
Opera (partial)

I like the `caniuse` layout https://caniuse.com/#search=web%20components
Doesn't have to be as advanced but some additionally browsers should be included.

tricki:
closes #499


adamdbradley:
thanks!

PR #608, opened 2020-05-06
docs(Custom Docs): Add Custom Docs Data Model

arayik-yervandyan:
Extend custom docs, add a description for JsonDocs and for JsonDocsComponent. Described common properties that are in the output of stencil custom docs generation.


adamdbradley:
this is great, thanks!

PR #618, opened 2020-05-08
docs: remove `componentDidUnload`

tricki:
Use of `componentDidUnload` is discouraged. It will eventually be removed and it's recommended to use `disconnectedCallback`

@adamdbradley:
> componentDidUnload() is difficult to be accurate 100% of the time since frameworks will sometimes reuse the same "element" it created in the dom, so it's hard for stencil to know if it's a new instance of a component, or that the element was just reused by a framework

### DIFF
--- a/src/components/custom-clock.tsx
+++ b/src/components/custom-clock.tsx
@@ -4,18 +4,19 @@ import { Component, State, h } from '@stencil/core';
   tag: 'custom-clock'
 })
 export class CustomClock {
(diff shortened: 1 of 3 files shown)


tricki:
Please check the order of the lifecylce methods in `style-guide.md`. I ordered them so `WillXYZ` comes before `DidXYZ` but the fact that some can be called multiple times makes ordering non-trivial.

PR #621, opened 2020-05-08
docs(resources): tabs don't have focus styling

tricki:
Added focus styling for easier keyboard navigation.


adamdbradley:
thanks

PR #625, opened 2020-05-10
Resources: add templates & tools

bitflower:
As discussed on Slack a new category "tools" would make sense. I took this approach a bit further and made the 3rd party category a pure "components" one. The official starters and all tools are now part of the new category.

Hope it makes sense?


adamdbradley:
awesome! Looks great

PR #628, opened 2020-05-11
Updated the default Shadow DOM is on by default, not off

manueledones:
In the stenciljs components generated, the default shadow DOM is on, the docs says it should be off


adamdbradley:
I think this is a little confusing yet. If you do not include the `shadow` property at all, it's sayin the default is that it's not using shadow dom. Bt you may be referring to how some of the starters are using `shadow: true`.

manueledones:
Oh now I get it :)

PR #631, opened 2020-05-16
[Resources-page] Adds link to WebComponents.dev

georges-gomes:
Points to a Stencil collection with the `My First Component` from the user guide and our basic `Counter` starter kit.
User can edit the components right away and there is minimal branding.


adamdbradley:
🚀

PR #632, opened 2020-05-19
docs(events): use type-safe events

noherczeg:
This is a more type-safe way of handling events. Helps writing better tests as well.


adamdbradley:
awesome thanks

PR #638, opened 2020-06-01
Update module-bundling.md

wbern:
Google leads many users to this page, and it's not completely apparent that you can actually code-split and dynamic import per component/module. This PR fixes that in the documentation.


adamdbradley:
Not sure this is the correct way to say it though. Stencil is very much so code splitting and generating files which will be lazy loaded, and it's doing this all automatically by already understanding how modules are used together. The `bundles` property is forcing certain components to be bundled together, but it's how lazy loading works in stencil. We can update the docs to try to state that.

PR #640, opened 2020-06-10
Update browser support

samber:
Ref:
- https://caniuse.com/#feat=custom-elementsv1
- https://caniuse.com/#feat=shadowdomv1


adamdbradley:
This may be misleading and makes it look like it doesn't work on modern versions of Edge. But it does work on Edge 16 to 18, and Edge 79+, so I think saying Edge 16+ works, otherwise Edge <= 18 looks like it doesn't support Edge 79+

PR #641, opened 2020-06-17
Fix url to DeviceDescriptors.js

dmartinjs:
the device list for puppeteer moved 
from https://github.com/GoogleChrome/puppeteer/blob/master/DeviceDescriptors.js
to https://github.com/puppeteer/puppeteer/blob/main/src/DeviceDescriptors.ts


adamdbradley:
thanks!

PR #645, opened 2020-06-19
docs(ssg): fix wrong method name

tricki:
Returning a Promise from `componentDidLoad` does not impact rendering.


adamdbradley:
Good catch, thanks

PR #651, opened 2020-07-06
docs(publishing): add existing publishing

tricki:
The page was created in f149dfd but not added to the readme and therefore not accessible.


adamdbradley:
thanks!

PR #655, opened 2020-07-08
Fix search menu + code blocks

claviska:
When the search menu is open, code blocks appear on top of the menu. This PR adds a small z-index to the site-menu tp address that.

I believe @fvaldes33 worked on the search feature, so he may be able to determine if this is the most appropriate way to fix this.

Before:
CleanShot 2020-07-08 at 09 29 00@2x

After:
CleanShot 2020-07-08 at 09 28 43@2x


fvaldes33:
@claviska yeah I think that should be fine. Thanks for catching that.

adamdbradley:
thanks!

PR #667, opened 2020-07-27
docs(ui): Reduce sidebar size

biirus:
As I mentioned in the issue, the main content section is too narrow right now. 
Screen Shot 2020-07-27 at 19 00 18

It's hard to read code blocks and text lines are too short (kinda newspapers style 😄 ). So, I reduced a sidebar width.
As a side-effect the `dist-custom-elements-bundle` nav item doesn't fit to the container once it's active (bold font increases the size). That's why I prevented word-wrapping and added `ellipsis` to nav elements.

The result:
Screen Shot 2020-07-27 at 19 00 03

### DIFF
--- a/src/components/search-bar/search-bar.css
+++ b/src/components/search-bar/search-bar.css
@@ -23,7 +23,7 @@ search-bar input {
 search-bar input {
   transition: border 0.3s;
   margin-right: 8px;
-  padding: 10px 16px 10px 56px;
+  padding: 10px 16px 10px 52px;
   font-size: 16px;
   border: 1px solid #DEE3EA;
   box-sizing: border-box;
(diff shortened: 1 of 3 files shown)


src/components/search-bar/search-bar.css
```diff
@@ -23,7 +23,7 @@ search-bar input {
 search-bar input {
   transition: border 0.3s;
   margin-right: 8px;
-  padding: 10px 16px 10px 56px;
+  padding: 10px 16px 10px 52px;
```
biirus, line 26: The left padding is not divisible by 8. It's made intentionally, cuz the search icon is not symmetric.
It means that focal mass is shifted to the left, so it would look not very accurate if we would have an equal distance from the left and right sides of the icon.

adamdbradley:
great update, thanks!

PR #670, opened 2020-08-05
Fixed a typo at the Rendering State section

aiyk:
"since these is the method" corrected to "since this is the method". These is referencing multiple methods, but the writers intention was to reference componentWillRender() only


adamdbradley:
thx

PR #673, opened 2020-08-17
Making the documentation for componentShouldUpdate() more explicit about it utility

nedredmond:
The documentation for this callback does not explain how to use it. Personally, I did not realize that I could provide arguments in order to get the scheduled updates until I saw the unit tests. From the ambiguous current description, it reads to me like an internal function that can't be intercepted.

### DIFF
--- a/src/docs/components/component-lifecycle.md
+++ b/src/docs/components/component-lifecycle.md
@@ -63,14 +63,15 @@ This hook is called when a component's `Prop` or `State` property changes and a 
 
 A couple of things to notice is that this method will not be executed before the initial render, that is, when the component is first attached to the dom, nor when a rerender is already scheduled in the next frame.
 
-Let’s say the following two props of a component change synchronously:
+The callback can be used to log scheduled updates:
 
 ```tsx
-component.somePropA = 42;
-component.somePropB = 88;
+componentShouldUpdate(newValue: any, oldValue: any, propName: string) {
+    console.log(`${propname} was updated\n` +
+        `New value: %&{newValue}\n` +
+        `Old value: %&{oldValue}`};
+}
 ```
-
-The `componentShouldUpdate` will be first called with arguments: `42`, `undefined` and `somePropA`. If it does return `true`, the hook will not be called again since the rerender is already scheduled to happen. Instead, if the first hook returned `false`, then `componentShouldUpdate` will be called again with `88`, `undefined` and `somePropB` as arguments, triggered by the `component.somePropB = 88` mutation.
 
 Since the execution of this hook might be conditioned, it's not good to rely on it to watch for prop changes, instead use the `@Watch` decorator for that.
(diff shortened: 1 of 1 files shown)


bodinaren:
I don't think this example represents the point of the function, and you also removed the explanation of how to actually use it. What you explain here is what the `@Watch()` decorator should be used for, and far more reliably. For example if two values changes at the same time and you do not return `false` in the first change, the `componentShouldUpdate` will not be called for the second change. A watch on each of those variables however will be triggered for both.

claviska:
> The `componentShouldUpdate` will be first called with arguments: `42`, `undefined` and `somePropA`. If it does return `true`, the hook will not be called again since the rerender is already scheduled to happen.

This is a critical piece of info that emphasizes why one _shouldn't_ use this method to watch props (which is a common misconception). It can probably be reworded, but I wouldn't suggest removing it entirely.

nedredmond:
@bodinaren @claviska 
I may be missing something here about this callback, and I would love it if you could fill me in. The original description advised when NOT to use it, but provided no information about when or how TO use it. I was successfully using this callback to debug errant re-renders during development. What other utility does it have? And if, per the original description, we are not supposed to use it, why is it exposed in the first place?

claviska:
The `componentShouldUpdate` method is used to let the developer decide if a rerender should occur. For example, if a prop changes but it doesn’t require a rerender, we can return `false` to tell Stencil to safely skip the render.

However, if `true` is returned, subsequent calls to the `componentShouldUpdate` method in the same render cycle will be skipped because Stencil is already aware that a rerender is needed.

This is why there's a warning to _not_ use the method for watching props. If `propA` and `propB` change in the same cycle and `true` is returned for `propA`, the method won't be called for `propB`.

Ultimately, this method serves as a way to opt-out of rendering when we know a prop change doesn't require it.

claviska:
> The callback can be used to log scheduled updates

This is an interesting way to use the method, but I don't think it belongs on this page since it conflicts with the method's intended usage. For example, if you returned `true` for any prop, you wouldn't see the logs for subsequent props.

You can achieve the same logging more reliably with `@Watch`, which is [demonstrated here](https://stenciljs.com/docs/reactive-data#watch-decorator) in the docs.

```js
@Watch('someProp')
watchHandler(newValue, oldValue) {
  console.log('The value of someProp changed: ', oldValue, newValue);
}
```

Re-reading [the docs for `componentShouldUpdate`](https://stenciljs.com/docs/component-lifecycle#componentshouldupdate), I agree that it's purpose is a bit ambiguous. I think adding a note about optimization, maybe something similar to my previous comment, would make things clearer.

claviska:
If there aren't any objections, I'm going to close this issue. If anyone has an idea to explain this method more clearly, let's start a fresh PR and go from there. 😄

PR #674, opened 2020-08-17
docs(templating): add warning for `key`

bodinaren:
I've seen several people that first hear about the `key` property use the index-parameter of the `map` function as the key like this:

```{ this.aList.map((item, idx) =>
  <div key={ idx }></div>
}```

This proposal adds a warning that this parameter is unfit as a `key` and explains why.


adamdbradley:
Good update, thanks

PR #679, opened 2020-08-31
Fix typo

BerkeAras:


BerkeAras:
> Thanks for catching this!

No problem 😄

PR #684, opened 2020-09-04
docs: adding components to 3rd party resource list

ftonato:


adamdbradley:
Svelte is great, but not sure a svelte debugger components needs to be listed in the stenciljs resources page. Is it ok we you remove that so the resources page can stay focused on stencil? Thanks

ftonato:
@adamdbradley Yeah, your suggestion make total sense. Do you have any suggestion where could I put this "debugger"?

PR #705, opened 2020-12-15
Remove reference to stencil_appload event in extras config

itsravenous:
This event has been renamed to `appload`, and made to be fired regardless of the value of `lifecycleDOMEvents` (see https://github.com/ionic-team/stencil/commits/346736)

I'll leave it to the  Stencil team to decide where best to document the `appload` event itself.


claviska:
Makes sense. I'm finding a better place to document this.

PR #715, opened 2021-01-28
Add Template to NewSpecPage Docs

say25:
## Work Done

- Adds Template and documentation about it

## Notes to Reviewer
- There are more options that are not captured in this document, I'm guessing these are a little out of date
- I've found `Template` particularly helpful when spec testing elements with their props, I don't mind adding an example of that to the docs as well.


Jbernardyoobic:
Hello @say25, I'm currently working on specs and I saw the template property. Do you have an example to show me ?

say25:
@Jbernardyoobic You can use it just like this, it allows you to pass in properties at page creation vs after page creation and requiring another render cycle

## Template Approach

```tsx
  const page = await newSpecPage({
    components: [MyComponent],
    template: () => <my-component id={mockId} />,
  });
```


## HTML Approach

```tsx
  const page = await newSpecPage({
    components: [MyComponent],
    html: '<my-component />',
  });
  
  page.rootInstance.id = mockId;
  await page.waitForChanges():
```

PR #725, opened 2021-03-31
Replace Vue docs URL string with hyperlink

arpi17:
This proposed change adds a hyperlink for faster access to the Vue docs instead of displaying the URL string


dotNetkow:
thank you!

PR #729, opened 2021-04-22
Update unit-testing.md

Tallyb:
Adding example on testing a component with the template syntax.

### DIFF
--- a/src/docs/testing/unit-testing.md
+++ b/src/docs/testing/unit-testing.md
@@ -29,10 +29,10 @@ import { MyCmp } from '../my-cmp';
 it('should render my component', async () => {
   const page = await newSpecPage({
     components: [MyCmp],
-    html: `<my-cmp title="Hello World"></my-cmp>`,
+    html: `<my-cmp></my-cmp>`,
   });
   expect(page.root).toEqualHtml(`
-    <my-cmp>Hello World</my-cmp>
+    <my-cmp>Success!</my-cmp>
   `);
 });
 ```
--- a/src/docs/testing/unit-testing.md
+++ b/src/docs/testing/unit-testing.md
@@ -38,7 +38,7 @@ it('should render my component', async () => {
 ```
 
 The example below uses the template option to test the component
-```
+```tsx
 // mycmp.spec.tsx
 // This is using jsx syntax and should be in a tsx file.
 import { h } from '@stencil/core';
--- a/src/docs/testing/unit-testing.md
+++ b/src/docs/testing/unit-testing.md
@@ -40,7 +40,7 @@ it('should render my component', async () => {
 The example below uses the template option to test the component
 ```tsx
 // mycmp.spec.tsx
-// This is using jsx syntax and should be in a tsx file.
+// Since the 'template' argument to `newSpecPage` is using jsx syntax, this should be in a .tsx file.
 import { h } from '@stencil/core';
 import { newSpecPage } from '@stencil/core/testing';
 import { MyCmp } from '../my-cmp';
--- a/src/docs/testing/unit-testing.md
+++ b/src/docs/testing/unit-testing.md
@@ -46,10 +46,10 @@ import { newSpecPage } from '@stencil/core/testing';
 import { MyCmp } from '../my-cmp';
 
 it('should render my component', async () => {
-  const title = 'Hello World';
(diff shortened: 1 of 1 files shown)


src/docs/testing/unit-testing.md
```diff
@@ -29,14 +29,34 @@ import { MyCmp } from '../my-cmp';
...
   });
   expect(page.root).toEqualHtml(`
-    <my-cmp>Success!</my-cmp>
+    <my-cmp>Hello World</my-cmp>
   `);
 });
 ```
 
+The example below uses the template option to test the component
+```
```
rwaskiewicz, line 41: Could we please update this line to denote this is a tsx file? That should help give some rendering engines an idea as to specific keywords it can highlight (similar to the snippet starting on line 25)

See how the top snippet is highlighted, and the bottom one is in all blue in GitHub's dark view:
Screen Shot 2021-07-15 at 7 04 19 PM

To do so, I think we just need to add `tsx` to this line like so:
```suggestion
```tsx
```
```

File: src/docs/testing/unit-testing.md [hunk 2]
```diff
@@ -29,14 +29,34 @@ import { MyCmp } from '../my-cmp';
...
+// mycmp.spec.tsx
+// This is using jsx syntax and should be in a tsx file.
```
rwaskiewicz, line 43: No action required, but this comment be implied by the comment above

src/docs/testing/unit-testing.md
```diff
@@ -29,14 +29,34 @@ import { MyCmp } from '../my-cmp';
 it('should render my component', async () => {
   const page = await newSpecPage({
     components: [MyCmp],
-    html: `<my-cmp></my-cmp>`,
+    html: `<my-cmp title="Hello World"></my-cmp>`,
   });
   expect(page.root).toEqualHtml(`
-    <my-cmp>Success!</my-cmp>
+    <my-cmp>Hello World</my-cmp>
```
rwaskiewicz, line 35: If we change this assertion, we should also update the [text above it](https://github.com/ionic-team/stencil-site/pull/729/files#diff-bc68864585479eef3eceee2c25bf4efcd68df4981723ee307e7e06f715abe7b0R22):

> In this example, when the component `MyCmp` renders it sets its text content as "Success!". The matcher `toEqualHtml()` is then used to ensure the component renders as expected.

Alternatively, we could also roll the changes back on lines 32 & 35

src/docs/testing/unit-testing.md
```diff
@@ -29,14 +29,34 @@ import { MyCmp } from '../my-cmp';
...
+// This is using jsx syntax and should be in a tsx file.
+import { h } from '@stencil/core';
+import { newSpecPage } from '@stencil/core/testing';
+import { MyCmp } from '../my-cmp';
+
+it('should render my component', async () => {
+  const title = 'Hello World';
+  const page = await newSpecPage({
+    components: [MyCmp],
+    template: () => (<my-cmp title={title}></my-cmp>),
```
rwaskiewicz, line 52: If someone were to use `title` as a `@Prop`, they'd get a warning like:

>  [ WARN  ]  Build Warn: ./src/components/my-cmp.tsx:12:11
           The @Prop() name "title" is a reserved public name. Please rename the
           "title" prop so it does not conflict with an existing standardized prototype
           member. Reusing prop names that are already defined on the element's
           prototype may cause unexpected runtime errors or user-interface issues on
           various browsers, so it's best to avoid them entirely.

It may be helpful to folks that aren't used to this warning to rename the attribute here something like 'greeting' or 'message':
```suggestion
  const greeting = 'Hello World';
  const page = await newSpecPage({
    components: [MyCmp],
    template: () => (<my-cmp greeting={greeting}></my-cmp>),
```

rwaskiewicz (commented):
Thanks @Tallyb! I had a few minor comments, please take a look at them and let me know what you think!

Tallyb:
Hi Ryan
You are welcome to adjust the PR as needed. My main issue was the lack of documentation. 
I explicitly added the tsx comment as it is not clear for people who wrote the tests prior to this change (tests were in ts files). 
Changing props names is absolutely fine.

rwaskiewicz (approved):
Thanks @Tallyb! I made a few of those tweaks - this definitely helps to have more testing docs 🙂

PR #737, opened 2021-05-29
fix: selection style and a11y

peterpeterparker:
**Current behavior:**

There are currently several issues with the `selection` style of the [Stencil](https://stenciljs.com/) site.

- **ignored** by Chrome and Safari: the defined style is ignored by these two browsers. Root cause of the issue is that both `selection` and `-moz-selection` are defined after each other as explained in this CSS-tricks [article](https://css-tricks.com/almanac/selectors/s/selection/).

> When browsers find a part of a select they don’t understand, they drop the entire thing, so this will fail all the time.

- **accessibility**: the selection style background is the same as the `code` background therefore, when applied as currently in Firefox, user cannot notice the difference when such elements are selected

**Solution:**

This PR fix the selection to be applied to all browsers and define another background color (and contrast).

**Alternative:**

Another alternative would be to simply remove the `selection` style, if you rather like this option let me know.

**Current behavior screenshots:**

Chrome:

Capture d’écran 2021-05-29 à 10 46 14

Safari:

Capture d’écran 2021-05-29 à 10 46 43

Firefox:

Capture d’écran 2021-05-29 à 10 47 16

**Solution screenshots:**

Chrome:

Capture d’écran 2021-05-29 à 10 49 15

Safari:

Capture d’écran 2021-05-29 à 10 49 24

Firefox:

Capture d’écran 2021-05-29 à 10 49 31


rwaskiewicz (approved):
Looks good! Thank you!!

peterpeterparker:
Coolio, thx Ryan!

PR #741, opened 2021-06-09
Update what-is-design-system.md

sdhandapani:
instead of Increased, it should be reduced/decreased time to market


src/docs/design-systems/what-is-design-system.md
```diff
@@ -25,6 +25,6 @@ A cohesive company brand strengthens its value in the minds of the customer. In
 By leveraging the components, patterns, and visual language of the Design System, designers can create design artifacts representing UI workflows. Developers refer to the artifacts as guidance for implementing the design with code.
 
 ## The Value of Design Systems
-With a design system in place, its true value is revealed. The entire product development team is freed up to focus on what matters most: solving customer problems and delivering value. Additionally, the avoidance of having teams working in parallel, recreating the same UI elements over and over,   has a real-world project impact in terms of increased time to market and cost savings.
+With a design system in place, its true value is revealed. The entire product development team is freed up to focus on what matters most: solving customer problems and delivering value. Additionally, the avoidance of having teams working in parallel, recreating the same UI elements over and over,   has a real-world project impact in terms of reduced time to market and cost savings.
```
rwaskiewicz, line 28: Good catch! I think the use of 'increased' was in reference to the cost savings, maybe it makes sense to add that adjective back like so:
```suggestion
With a design system in place, its true value is revealed. The entire product development team is freed up to focus on what matters most: solving customer problems and delivering value. Additionally, the avoidance of having teams working in parallel, recreating the same UI elements over and over,   has a real-world project impact in terms of reduced time to market and increased cost savings.
```
(this keeps your suggestion to used 'reduced' as well)

rwaskiewicz (changes requested):

senthildhandapani (approved):
great, looks ok now

rwaskiewicz (approved):

PR #746, opened 2021-06-28
Clarification on using @Watch on @State

DodoTheDeveloper:
Was initially not clear found someone who had the same question on SO https://stackoverflow.com/questions/67717860/stenciljs-can-watch-be-used-on-state

### DIFF
--- a/src/docs/components/reactive-data.md
+++ b/src/docs/components/reactive-data.md
@@ -18,18 +18,24 @@ When a component updates because of a state change (props or state change), the 
 
 ## Watch Decorator
 
-When a user updates a property, `Watch` will fire the method it's attached to, and pass that method the new value of the prop along with the old value. `Watch` is useful for validating props or handling side effects. The `Watch` decorator does not fire when a component initially loads.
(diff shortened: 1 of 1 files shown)


src/docs/components/reactive-data.md
```diff
@@ -18,19 +18,25 @@ When a component updates because of a state change (props or state change), the
 
 ## Watch Decorator
 
-When a user updates a property, `Watch` will fire the method it's attached to, and pass that method the new value of the prop along with the old value. `Watch` is useful for validating props or handling side effects. The `Watch` decorator does not fire when a component initially loads.
+When a user updates a property, `Watch` will fire the method it's attached to, and pass that method the new value of the prop along with the old value. `Watch` is useful for validating props, states or handling side effects. The `Watch` decorator does not fire when a component initially loads.
```
rwaskiewicz, line 21: Thanks for this! This led to a great discussion amongst the team about some of the ways our documentation is lacking, and we think this definitely helps! To help make this even clearer for folks that don't find that StackOverflow post you mentioned, what do you think of this verbiage?

```suggestion
`Watch` will fire the method it's attached to when a user updates a property, or when an internal state member changes. That method will receive the new value of the prop/state, along with the old value. `Watch` is useful for validation or the handling of side effects. The `Watch` decorator does not fire when a component initially loads.
```

rwaskiewicz (changes requested):

rwaskiewicz (approved):

PR #761, opened 2021-08-24
Update config to document invisiblePrehydration

splitinfinities:
Merge when https://github.com/ionic-team/stencil/pull/2989 is merged.

### DIFF
--- a/src/docs/config/overview.md
+++ b/src/docs/config/overview.md
@@ -105,7 +105,7 @@ Check out the [styling docs](https://stenciljs.com/docs/styling#global-styles) o
 When true, `invisiblePrehydration` will use Ionic's opinion to visually hide components before they are hydrated by adding an automatically injected style tag to the document's head. Setting `invisiblePrehydration` to `false` will not inject the style tag into the head, allowing you to style your web components pre-hydration. Please note that setting `invisiblePrehydration` to false will cause everything to be visible when your page is loaded, causing more prominent Flash of Unstyled Content (FOUC), however you can style your web component's fallback content to your preference.
 
 ```tsx
-invisiblePrehydration: boolean
+invisiblePrehydration: true
 ```
 
 ## hashFileNames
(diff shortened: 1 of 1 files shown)


src/docs/config/overview.md
```diff
@@ -97,6 +98,13 @@ globalStyle: 'src/global/app.css'
 
 Check out the [styling docs](https://stenciljs.com/docs/styling#global-styles) of how to use global styles in your app.
 
+## invisiblePrehydration
+
```
rwaskiewicz, line 102: [Other fields have the default between the name of the field and the description](https://github.com/ionic-team/stencil-site/pull/761/files#diff-689c19470ae2ed5ac0b4ba2bb534871dd960e99ae1638ca7da6bab934a5dd022R122). Could we do something like that here?

```suggestion

*default: `true`*

```

src/docs/config/overview.md
```diff
@@ -97,6 +98,13 @@ globalStyle: 'src/global/app.css'
...
+Sets if the build should add an automatically injected style tag to the document's head. It defaults to `true` and will use Ionic's opinion to visually hide components while they are not hydrated. Setting `false` will let you style your web components pre-hydration. Note that this will cause everything to be visible on load, and you will be able to style your fallback content how you would perfer.
```
rwaskiewicz, line 103: Minor, if we accept my comment above, we may want to reword this a bit. One way (but definitely not the only way!):

```
When true, `invisibleHydration` will use Ionic's opinion to visually hide components before they are hydrated. Setting `false` will let you style your web components pre-hydration. Note that this will cause everything to be visible on load, and you will be able to style your fallback content how you would prefer. 
```

rwaskiewicz, line 103: Regardless of the reword, the last word in this section has a typo:
`perfer` -> `prefer`

rwaskiewicz (changes requested):

PR #766, opened 2021-09-01
refactor(docs): corrected some links

mfranzke:
### Summary of changes

- The repository https://github.com/jepiqueau/stencil-flip-images doesn't exist any more, and there wasn't a replacement to find during a quick research, so I've removed the link in total
- The blog entry "How Lazy-Loading Web Components Work with Stencil" ([/blog/how-lazy-loading-web-components-work](https://ionicframework.com/blog/how-lazy-loading-web-components-work)) doesn't exist any more – it must have been either removed intentionally or hasn't been migrated to the new home of the blog entries on https://ionicframework.com/. So I've added an archived version of that page as the link target as I still expect that content to be relevant.
- The `master` branch of the https://github.com/ionic-team/ionic/ repository has been renamed to `main`, so we had to change the related link to prevent the redirect.
- The link to https://github.com/ionic-team/stencil/blob/master/LICENSE leads to a 404 error, as the fileending was missing.
- https://facebook.github.io/react/docs/introducing-jsx.html redirects to https://reactjs.org/docs/introducing-jsx.html so we could insert that link target directly.


rwaskiewicz (approved):
Looks great! Thanks!

PR #767, opened 2021-09-03
docs(config): add documentation for scopedSlotTextContentFix

rwaskiewicz:
Add documentation for https://github.com/ionic-team/stencil/pull/3047

### DIFF
--- a/src/docs/config/extras.md
+++ b/src/docs/config/extras.md
@@ -76,7 +76,7 @@ Safari 10 supports ES modules with `<script type="module">`, however, it did not
 
 ### scopedSlotTextContentFix
 
-An experimental flag that when set to `true`, aligns the behavior of invoking the `textContent` getter/setter on a scoped component to act more like a component that uses the shadow DOM. Specifically, invoking `textContent` on a component will adhere to the return values described in the [DOM Spec](https://dom.spec.whatwg.org/#dom-node-textcontent). Defaults to `false`.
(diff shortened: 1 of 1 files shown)


src/docs/config/extras.md
```diff
@@ -74,6 +74,10 @@ Dispatches component lifecycle events. By default these events are not dispatche
 
 Safari 10 supports ES modules with `<script type="module">`, however, it did not implement `<script nomodule>`. When set `safari10` is set to `false`, the runtime will not patch support for Safari 10. If the app does not need to support Safari 10, it's recommended to set this to `false`. Defaults to `false`.
 
+### scopedSlotTextContentFix
+
+An experimental flag that when set to `true`, aligns the behavior of invoking the `textContent` getter/setter on a scoped component to act more like a component that uses the shadow DOM. Defaults to `false`.
```
a-giuliano, line 79: How does it behave that makes it more like a component that uses the shadow DOM? This may be unclear if the reader doesn't know what the behavior for a component using shadow DOM is.

rwaskiewicz: I updated the verbiage to point to the spec. That may not be the most user friendly thing TBH, but spelling out every single scenario would be pretty cumbersome. Alternatively, we could link https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#description, but I don't know if one should be considered "better" than the other - the MDN one is definitely more user friendly, but the spec is, well, the spec 😛 LMK what you think

a-giuliano: Yeah tough call, but I think I would lean more towards MDN. At the end of the day, the docs are meant to be informative and understandable/user friendly, so as long as MDN provides all the necessary info (and I believe it does) I would choose it for its user friendliness.

rwaskiewicz: 👍 Done in https://github.com/ionic-team/stencil-site/pull/767/commits/4d8076c10030f64ec2a0a656a729e881ed385d42

a-giuliano (approved):

splitinfinities (approved):

ltm (approved):

PR #776, opened 2021-09-28
docs(preamble): rework preamble documentaiton

rwaskiewicz:
reword the preamble explainer to be more clear regarding 'which' files
receive the generated preamble

specify the premable will result in a pinned comment

specify how newlines are honored

provide an example to demonstrate the above two points


src/docs/config/overview.md
```diff
@@ -160,12 +161,22 @@ Please see the [Dev-Server docs](/docs/dev-server).
 
 *default: `undefined`*
 
-The `preamble` configuration is a `string` that represents a preamble in the main file of the build. Help to persist a banner or add relevant information about the resulting build.
+Used to help to persist a banner or add relevant information about the resulting build, the `preamble` configuration 
+field is a `string` that will be converted into a pinned comment and placed at the top of all emitted JavaScript files,
```
splitinfinities, line 165: Something I would possibly note here is that the preamble doesn't get placed on resultant framework binding files as of today.

splitinfinities: Not sure how best to word this today though... Possibly as a note underneath the example? Or maybe it's not worth calling out quite yet. I don't mind - your call.

src/docs/config/overview.md
```diff
@@ -160,12 +161,22 @@ Please see the [Dev-Server docs](/docs/dev-server).
...
+with the exception of any emitted polyfills. Escaped newlines may be placed in the provided value for this field and 
+will be honored by Stencil.
 
+Example:
 ```tsx
-preamble: 'Built with Stencil'
+preamble: 'Built with Stencil\nCopyright (c) SomeCompanyInc.'
```
splitinfinities, line 171: Just for my knowledge, do template string literals maintain there spacing here too?

splitinfinities (approved):
Lookin' good! Only a couple higher level thoughts that may or may not be worth considering.

src/docs/config/overview.md
rwaskiewicz: I'd lean omitting it for now, but a great point!

PR #784, opened 2021-11-01
docs(props): add documentation for exporting prop types

rwaskiewicz:
add documentation around types used by @props and their visibility

STENCIL-222: bug: @Prop can't work with typescript's Type alias

### DIFF
--- a/src/docs/components/properties.md
+++ b/src/docs/components/properties.md
@@ -25,8 +25,7 @@ export class TodoList {
 ```
 
 When using user-defined types like `MyHttpService`, the type must be exported using the `export` keyword. Above,
-`MyHttpService` is imported from `'../some/local/directory/MyHttpService'`, and must be exported from that file 
-(otherwise we'd have no way to use it).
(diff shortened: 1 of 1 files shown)


src/docs/components/properties.md
```diff
@@ -4,15 +4,41 @@ description: Properties
...
+  @Prop() color: string;
+  @Prop() favoriteNumber: number;
+  @Prop() isSelected: boolean;
+  @Prop() myHttpService: MyHttpService;
+}
+```
+
+When using user-defined types like `MyHttpService`, the type must be exported using the `export` keyword. Above,
+`MyHttpService` is imported from `'../some/local/directory/MyHttpService'`, and must be exported from that file 
+(otherwise we'd have no way to use it).
```
a-giuliano, line 29: I don't think you need the line in parens. I think your explanation is good.

rwaskiewicz: Done! https://github.com/ionic-team/stencil-site/pull/784/commits/bec3cc5b3159cf1277aa5af7b2f87aaa4bacbb5a

splitinfinities (approved):
This rocks, thank you so much!

PR #802, opened 2021-12-06
Update Angular docs to consider the NgModule

splitinfinities:
The Angular component library needs to have an NgModule, and not just the components


rwaskiewicz:
After discussing with the Framework and Stencil teams, we've decided to close this. We're going to be investigating the value-add of something more along the lines of Angular's standalone components, as well as how to best steer folks in the right direction depending on whether they're using the `lazy` output target vs `dist-custom-elements`.

PR #803, opened 2021-12-07
docs(angular): replace `React` with `Angular`

dgonzalezr:
- The documentation for Angular framework integration mentions `React` in some places where it should be `Angular` instead.


rwaskiewicz (approved):
Thanks!

PR #804, opened 2021-12-16
docs(testing): add explicit mention of testing lib versions

rwaskiewicz:
# Changes

state the versions of jest and puppeteer that are supported by stencil. this is the 'required' work for supporting Jest 27+. however, i took an initial stab at some rework of the testing overview page, but by no means is it "good enough" for us to mark it off our sheet of pages we want to improve. the test configuration & IDE integration could certainly use a little more love, but I think it's in a slightly better state than where it was/is today


a-giuliano:
This is great. We'll certainly take another pass at it as part of our docs campaign, but I really like the improvements here.

a-giuliano (approved):

PR #805, opened 2021-12-23
Fix incorrect name of Component Lifecycle Method

eusougz:
❌ Incorrect name: `componentShouldRender`
✅ Correct name: `componentShouldUpdate`

### DIFF
--- a/src/docs/guides/style-guide.md
+++ b/src/docs/guides/style-guide.md
@@ -213,7 +213,7 @@ export class Something {
   componentWillUpdate() {}
   componentDidUpdate() {}
   componentWillRender() {}
-  componentShouldRender(newVal: any, oldVal: any, propName: string) {}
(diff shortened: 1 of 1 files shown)


src/docs/guides/style-guide.md
```diff
@@ -213,7 +213,7 @@ export class Something {
   componentWillUpdate() {}
   componentDidUpdate() {}
   componentWillRender() {}
-  componentShouldRender(newVal: any, oldVal: any, propName: string) {}
+  componentShouldUpdate(newVal: any, oldVal: any, propName: string) {}
```
rwaskiewicz, line 216: Thanks! Since the style guide states (in the JSDoc above this section):

>   * Ordered by their natural call order, for example
>   * WillLoad should go before DidLoad.

Would you mind moving this line to come before `componentWillUpdate() {}`?

rwaskiewicz (changes requested):

PR #816, opened 2022-01-14
docs(react): update react wrapper docs

rwaskiewicz:
this pr makes a few minor adjustments to the react wrapper docs after running through them myself.  most notably, it:
- attempts to reinforce the React Component Library:
  - wraps Stencil components
  - holds our React wrapper components
- making the linking process a little easier by:
  - having folks clone the stencil ds react template to a directory whose name matches the template's `package.json#name` value
  - avoid having to switch directories by running `npm build` before `npm link`
- clarify how exporting components from the React Component Library works (mostly that it is not needed if you use the template)

### DIFF
--- a/src/docs/framework-integration/react.md
+++ b/src/docs/framework-integration/react.md
@@ -55,7 +55,7 @@ mv {the path to your Stencil component library} {the path to your monorepo}
 Next, we will need to create the React component library that will wrap your Stencil components. This library will be a sibling to your Stencil component library. Inside your monorepo, you can create your own React project, or you can use the [React component library template](https://github.com/ionic-team/stencil-ds-react-template) to bootstrap it. To do this, run the following command
 
 ```bash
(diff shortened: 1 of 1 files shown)


src/docs/framework-integration/react.md
```diff
@@ -127,66 +128,74 @@ You’ll see the new generated file in your React component library at the locat
...
+
+First, build your Stencil component library. In your **Stencil component library**, run
 
 ```bash
-npm link
+npm run build
 ```
 
-With the symlink created, we next need to specify which packages will be consuming them. Your React component library will need to consume your Stencil component library. In the directory of your **React component library** run
+Now you can create a symlink for your Stencil component library. From withing your **Stencil component library**, run
```
a-giuliano, line 159: typo: within

a-giuliano (approved):
Looks good. Just found one typo. I'll probably come around at some point to take another look at this, but in the interest of timing, I think this is good to go.

PR #819, opened 2022-01-20
wrong cd folder target after cloning react wrapper project

jsbaguette:
I think it's what you wanted to have here no ?


rwaskiewicz:
> I think it's what you wanted to have here no ?

Yep! Nice catch!

rwaskiewicz (approved):

PR #827, opened 2022-01-27
Incomplete documentation.

IndyVC:
@Watch decorates passes a third parameter, the name of the property that is changed. Very handy when you are using stencil-store and want to update the store accordingly when the initial prop/state changes.


rwaskiewicz (approved):
Thanks!

PR #830, opened 2022-02-02
add exportparts to styling docs

a-giuliano:


src/docs/components/styling.md
```diff
@@ -162,6 +162,45 @@ shadow-card::part(heading) {
...
+    return (
+      <Host>
+        <h1>Outer Component</h1>
+        <inner-component exportparts="inner-text" />
+      </Host>
+    );
+  }
+}
+
+export class InnerComponent {
```
rwaskiewicz, line 181: It may be worth adding a `@Component` decorator to this class, so that we define a tag to be associated with it. That way, [this line](https://github.com/ionic-team/stencil-site/pull/830/files#diff-fdd264bad2c75226b929f996da39e4dbee5ced97ea40f7c2eeb86ef9ea110650R175) makes a little more sense. (also consider adding `shadow` as well)

Perhaps we should add it (`@Component`) for both, since we reference `outer-component` [here](https://github.com/ionic-team/stencil-site/pull/830/files#diff-fdd264bad2c75226b929f996da39e4dbee5ced97ea40f7c2eeb86ef9ea110650R201)

src/docs/components/styling.md
```diff
@@ -162,6 +162,45 @@ shadow-card::part(heading) {
...
+    return (
+      <Host>
+        <h1 part="inner-text">Inner Component</h1>
+      </Host>
+    );
+  }
+}
+```
+
+By specifying "inner-text" as the value of the `exportparts` attribute, elements of the `InnerComponent` with a `part` of "inner-text" can now be sty;ed in the light DOM. Even though the `InnerComponent` is not used directly, we can style its parts through the `OuterComponent`.
```
rwaskiewicz, line 192: typo: 'sty;ed' -> 'styled'

```suggestion
By specifying "inner-text" as the value of the `exportparts` attribute, elements of the `InnerComponent` with a `part` of "inner-text" can now be styled in the light DOM. Even though the `InnerComponent` is not used directly, we can style its parts through the `OuterComponent`.
```

rwaskiewicz, line 192: Not necessary ATM, one thing I want us to start thinking about is consistency when referring to components (consistency throughout the doc site that is).  At some point in the near future, I want to decide if we should refer to components by their tag (`outer-component`) vs their class (`OuterComponent`). We can always deviate where necessary to get a particular point across - this is something that's cropped up on my end while working through the props docs

rwaskiewicz (changes requested):

src/docs/components/styling.md
a-giuliano: yeah I was thinking about this as well as I was writing it. Let's circle back to this

rwaskiewicz (approved):

PR #832, opened 2022-02-04
docs(props): rework props documentation

rwaskiewicz:
work in progress


src/docs/guides/typed-components.md
```diff
@@ -30,36 +30,5 @@ button.fill = 'outline';
 
 **IMPORTANT**: always use the `HTML{}Element` interfaces in order to hold references to components.
 
+TODO: Put a redirect/link here for the links I just deleted
```
rwaskiewicz, line 33: I need to get these redirects set up still

src/docs/components/properties.md
```diff
@@ -7,219 +7,592 @@ contributors:
...
 export class TodoList {
-  @Prop() color: string;
-  @Prop() favoriteNumber: number;
-  @Prop() isSelected: boolean;
-  @Prop() myHttpService: MyHttpService;
+    // Second, we decorate a class member with @Prop()
+    @Prop() name: string;
+    
+    render() {
+        // Within the a component's class, its props are
```
sean-perkins, line 31:
```suggestion
        // Within the component's class, its props are
```

src/docs/components/properties.md
```diff
@@ -7,219 +7,592 @@ contributors:
...
+
+Props are declared on a component using Stencil's `@Prop()` decorator, like so:
 
 ```tsx
-// TodoList.tsx
-import { Prop } from '@stencil/core';
-import { MyHttpService } from '../some/local/directory/MyHttpService';
-...
+// First, we import Prop from '@stencil/core'
+import { Component, Prop, h } from '@stencil/core';
```
sean-perkins, line 21: ~ Small detail, but other usage examples order `h` first. Likely all usage examples should be the same, either alphabetically or an explicit order.

sean-perkins (commented):
Small details, new documentation looks great. Love the extra detail 👏

src/docs/components/properties.md
rwaskiewicz: Good point, let me double check the Stencil CLI and match that

src/docs/guides/typed-components.md
rwaskiewicz: I went for a 'cheap' way to do this in https://github.com/ionic-team/stencil-site/pull/832/commits/9638d586e3da7209de33d8fecd983439aced85fc - I don't think Vercel supports redirects that use document fragments. So while I could get redirects up to the 'properties' page, it's not a great experience trying to find the section from this page that maps to the properties page.

src/docs/components/properties.md
```diff
@@ -7,219 +7,592 @@ contributors:
...
+// Set isComplete to 'false'
+<todo-list-item isComplete={false}></todo-list-item>
+```
 
-...
-export class NameElement {
+There are a few ways in which Stencil treats props that are of type `boolean` that are worth noting:
+
+1. The value of a boolean prop will be `false` if provided the string "false" in HTML
+    ```html
```
a-giuliano, line 198: Do you happen to know why these code snippets are rendering weird? I don't love the inner box.

src/docs/components/properties.md
```diff
@@ -7,219 +7,592 @@ contributors:
...
+    @Prop() aNumber = 42;
+    @Prop() aString = 'defaultValue';
+
+    render() {
+        return <div>The number is {this.aNumber} and the string is {this.aString}</div>
+    }
 }
 ```
+Regardless of if we use this component in HTML or TSX, "The number is 42 and the string is defaultValue" is displayed
+when no values are passed to out component:
```
a-giuliano, line 380:
```suggestion
when no values are passed to our component:
```

src/docs/components/properties.md
```diff
@@ -7,219 +7,592 @@ contributors:
...
+
+This component has **3 properties**, but the compiler will create **only 2 attributes**: `is-complete` and
+`thing-to-do`.
+
+```html
+<todo-list-item is-complete="false" thing-to-do="Read Attribute Naming Section of Stencil Docs"></my-cmp>
+```
+
+Notice that the `myHttpService` type is not a primitive (e.g. not a `number`, `boolean`, or `string`). Since DOM
+attributes can only be strings, it does not make sense to have an associated DOM attribute called "my-http-service".
```
a-giuliano, line 511: Might be helpful to show how you can still set the value of `myHttpService`.

src/docs/components/properties.md
```diff
@@ -7,219 +7,592 @@ contributors:
...
 
-## Prop options
+> Note: Children components should not know about or reference their parent components.  This allows Stencil to
+> efficiently re-render your components. Passing a reference to a component as a prop may cause unintended side effects.
 
-The `@Prop(opts?: PropOptions)` decorator accepts an optional argument to specify certain options, such as the `mutability`, the name of the DOM attribute or if the value of the property should or shouldn't be reflected into the DOM.
+## Mutability
+
+A Prop is by default immutable from inside the component logic. Once a value is set by a user, the component cannot
+update it internally.
```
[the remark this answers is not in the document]
rwaskiewicz: 👍  added in 7b6421bd0

src/docs/components/properties.md
rwaskiewicz: Nor do I. Unfortunately it's an artifact of the system we use to render code snippets, and even more unfortunately I don't think there's a way around it with our current infra 😢

src/docs/components/properties.md
rwaskiewicz: I was a little intentionally vague here to avoid scope creep - I stubbed `MyHttpService` as an empty class earlier in this doc, and didn't want to get into all the details as a part of this PR. How would you feel about us doing that as a follow on effort?

a-giuliano: I think that's totally fair. I'm notoriously bad at avoiding scope creep so I admire your discipline.

src/docs/components/properties.md
a-giuliano: Ugh that's so weird. When I changed the indentation it was fixed on a local build, so I added it here [375f115](https://github.com/ionic-team/stencil-site/pull/832/commits/375f11580f936c418175f8962dc403afd86df55f). Doesn't seem like it's fixed in the preview.

rwaskiewicz: Yeah, I couldn't get it to go away either 😆 😢 TBH I'm not too worried about it - it's not ideal, but I think we're gonna start the conversations around moving this site to Docusaurus, so hopefully it's a short lived problem

a-giuliano: Nvm. I forgot a newline. Fixed here [d87be29](https://github.com/ionic-team/stencil-site/pull/832/commits/d87be29ce55a48bc0884201bdaab43526f0afff9)

rwaskiewicz: Oh nice!!

a-giuliano (approved):

PR #834, opened 2022-02-10
docs(state): refactor state prop documentation 

rwaskiewicz:
this commit updates the documentation for the `@State()` decorator. the
goal of this commit is to make the documentation clearer around the
decorator's usage, while providing additional examples.

### DIFF
--- a/src/docs/components/state.md
+++ b/src/docs/components/state.md
@@ -9,25 +9,12 @@ contributors:
 
 # State
 
-'State' is a general term that refers to the values and objects that are stored on a class for use now or in the future.
-The TypeScript class `CurrentTime` below stores the time that it was created, and is able to print it back to the user:
-
-```ts
-class CurrentTime {
-    constructor(public creationTime: number) { }
-
-    whatTimeWasItCreated(): string {
-        return new Date(this.creationTime).toLocaleTimeString();
-    }
-}
-```
-
-Here, `creationTime` is a member of the `CurrentTime` class that stores the time an instance of the class was created.
-We say that `creationTime` is part of the _state_ of `CurrentTime`.
+'State' is a general term that refers to the values and objects that are stored on a class or an instance of a class for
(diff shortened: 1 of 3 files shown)


src/docs/components/state.md
```diff
@@ -4,73 +4,277 @@ description: Use the State() for component's internal state
...
 }
 ```
 
-## When to use?
+## Examples
 
-Not all internal state might need to be decorated with `@State()`, in fact it's a good practice to avoid using it if you know for sure that the value will not change or that it does NOT need to trigger a re-rendering:
+### Using `@State()` with `@Listen()`
+
+This example makes use of `@State` and [`#Listen`](/docs/events#listen-decorator) decorators. We define a class member
```
a-giuliano, line 168:
```suggestion
This example makes use of `@State` and [`Listen`](/docs/events#listen-decorator) decorators. We define a class member
```

src/docs/components/state.md
```diff
@@ -4,73 +4,277 @@ description: Use the State() for component's internal state
...
 
-# State Decorator
+# State
 
-The `@State()` decorator can be used to manage internal data for a component. This means that a user cannot modify this data from outside the component, but the component can modify it however it sees fit. Any changes to a `@State()` property will cause the components `render` function to be called again.
+'State' is a general term that refers to the values and objects that are stored on a class for use now or in the future.
+The TypeScript class `CurrentTime` below stores the time that it was created, and is able to print it back to the user:
 
-## Example
+```ts
```
a-giuliano, line 15: I'm not sure this example adds much. I think your explanation here is very good, but may be better suited in the next section (The State Decorator). For me, it makes more sense to say here's the state decorator and here's why/when we use it all in one section.

a-giuliano (commented):
Really great. One typo catch and an organizational point.

a-giuliano (approved):
Love it

PR #837, opened 2022-02-16
fix: fix typo

switchnollie:
There is a typo in the docs: It should be "JavaScript" instead of "JavaScrip"


rwaskiewicz (approved):
Thanks!

PR #840, opened 2022-02-23
docs(angular-framework-integration): set-up instructions

sean-perkins:
Updates the usage instructions for setting up the `@stencil/angular-output-target` as found here: https://stenciljs.com/docs/angular#what-is-the-best-format-to-write-event-names

- Fixes an incorrect implementation detail where Angular components were being added to the `imports` array (which is not correct for Angular modules).
- Uses `npx` to avoid installing global dependencies
- Removes npm vs. yarn syntax (developers are commonly aware)
- Uses less verbose language for instructions
- Specifies the advantages of using the component wrappers over the vanilla web components

Does not modify language of FAQ and API. Those sections should be updated at a later date.

### DIFF
--- a/src/docs/framework-integration/angular.md
+++ b/src/docs/framework-integration/angular.md
@@ -132,7 +132,7 @@ npm link
 From your Angular component library's directory, run the following command:
 
 ```bash
-npm link name-of-your-stencil-package`
+npm link name-of-your-stencil-package
 ```
 
 The name of your Stencil package should match the `name` property from the Stencil component library's `package.json`.
--- a/src/docs/framework-integration/angular.md
+++ b/src/docs/framework-integration/angular.md
@@ -16,7 +16,7 @@ contributors:
 
 **Supports: Angular 12+ • TypeScript 4.0+ • Stencil v2.9.0+**
 
-Stencil can generate Angular component wrappers for your web components. The benefits of the component wrappers over the standard web components are:
+Stencil can generate Angular component wrappers for your web components. This allows your Stencil components to be used within an Angular application.  The benefits of using Stencil's component wrappers over the standard web components include:
 
 - Angular component wrappers will be detached from change detection, preventing unnecessary repaints of your web component.
 - Web component events will be converted to RxJS observables to align with Angular's `@Output()` and will not emit across component boundaries.
--- a/src/docs/framework-integration/angular.md
+++ b/src/docs/framework-integration/angular.md
@@ -14,17 +14,21 @@ contributors:
 
 # Angular Integration
 
-**Support: __Angular 12+__ • __TypeScript 4.0+__ • __Stencil v2.9.0+__**
+**Supports: Angular 12+ • TypeScript 4.0+ • Stencil v2.9.0+**
 
-Stencil provides a wrapper for your custom elements to be used as first-class Angular components. The goal of a wrapper is to smooth over how Stencil’s code works within a framework. Wrappers provide a function that you can use within Stencil’s Output Targets to automatically create components for the targeted framework that wrap the web components you author in a Stencil project.
+Stencil can generate Angular component wrappers for your web components. The benefits of the component wrappers over the standard web components are:
 
-One benefit of the wrapper pattern includes improved maintainability since you are writing code once, and reusing it across frameworks. Another benefit of this pattern is that you can have first-class integration with your framework of choice. For example, with the Angular wrapper, you can bind input events directly to a value accessor for seamless integration in Angular’s bi-directional data flow. 
+- Angular component wrappers will be detached from change detection, preventing unnecessary repaints of your web component.
+- Web component events will be converted to RxJS observables to align with Angular's `@Output()` and will not emit across component boundaries.
+- _Optionally_ form control web components can be used as control value accessors with Angular's reactive forms or `[ngModel]`.
 
 ## Setup
 
 ### Project Structure
 
-To organize the generated component libraries for different frameworks, we recommend using a monorepo structure. This monorepo will contain your Stencil component library as well as the component libraries for whatever frameworks you choose. The overall structure of a monorepo with Stencil and Angular component libraries might look something like this:
+We recommend using a monorepo structure for your component library with component wrappers. Your project workspace should contain your Stencil component library and the library for the generated Angular component wrappers.
+
+An example project set-up may look similar to:
 
 ```
 top-most-directory/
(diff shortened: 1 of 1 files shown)


src/docs/framework-integration/angular.md
```diff
@@ -39,211 +43,179 @@ top-most-directory/
...
+```ts
+export { COMPONENTS } from './stencil-generated/components';
 ```
 
 ### Link your packages (optional)
 
-If you’re using a monorepo tool like Lerna or Nx, you can skip this step. Before you can successfully build a local version of your Angular component library, you will need to link the Stencil package to the Angular package.
-
-**First, in your Stencil directory, run the following command:**
+> If you are using a monorepo tool (Learn, Nx), skip this section.
```
a-giuliano, line 122:
```suggestion
> If you are using a monorepo tool (Lerna, Nx), skip this section.
```

src/docs/framework-integration/angular.md
```diff
@@ -39,211 +43,179 @@ top-most-directory/
...
+```bash
+npm link
+```
 
-Or
+From your Angular component library's directory, run the following command:
 
-`yarn link {the name of your Stencil package}`
+```bash
+npm link name-of-your-stencil-package`
```
a-giuliano, line 135:
```suggestion
npm link name-of-your-stencil-package
```

a-giuliano (approved):
Looks really good. Just caught two typos

src/docs/framework-integration/angular.md
```diff
@@ -39,211 +43,179 @@ top-most-directory/
...
+```
 
-`yarn build`
+If the build is successful, you will now have contents in the file specified in `directivesProxyFile`.
 
-Once the build is complete, you will see new files in your Angular component library's directory!
+You can now finally import and export the generated component wrappers for your component library. For example, in your main `AppModule` you may do the following:
 
+```ts
+import { COMPONENTS } from './stencil-generated/components';
```
rwaskiewicz, line 105: This may be a bit confusing to folks. I totally see what you're going for here - import the components from this path and add them to your declarations & exports fields.  However `./stencil-generated/components` also has this line:
```
import { Components } from 'stencil-library';
```
where I wonder if folks may think we've just made a typo with the casing of 'components' here. Does it make sense if we treat this guide more like a tutorial and give move concrete names here?
```
import { MyComponent } from './stencil-generated/components';

@NgModule({
  declarations: [
    ComponentLibraryComponent,
    MyComponent
  ],
  imports: [
  ],
  exports: [
    ComponentLibraryComponent
  ]
})
```

src/docs/framework-integration/angular.md
```diff
@@ -39,211 +43,179 @@ top-most-directory/
...
-Or
+```bash
+npm run build
+```
 
-`yarn build`
+If the build is successful, you will now have contents in the file specified in `directivesProxyFile`.
 
-Once the build is complete, you will see new files in your Angular component library's directory!
+You can now finally import and export the generated component wrappers for your component library. For example, in your main `AppModule` you may do the following:
```
rwaskiewicz, line 102: Should this be the main `AppModule` for our Angular application, or for our component library? I'm a little confused here

src/docs/framework-integration/angular.md
```diff
@@ -14,17 +14,21 @@ contributors:
...
+Stencil can generate Angular component wrappers for your web components. The benefits of the component wrappers over the standard web components are:
 
-One benefit of the wrapper pattern includes improved maintainability since you are writing code once, and reusing it across frameworks. Another benefit of this pattern is that you can have first-class integration with your framework of choice. For example, with the Angular wrapper, you can bind input events directly to a value accessor for seamless integration in Angular’s bi-directional data flow. 
+- Angular component wrappers will be detached from change detection, preventing unnecessary repaints of your web component.
+- Web component events will be converted to RxJS observables to align with Angular's `@Output()` and will not emit across component boundaries.
+- _Optionally_ form control web components can be used as control value accessors with Angular's reactive forms or `[ngModel]`.
```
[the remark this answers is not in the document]
sean-perkins: Makes sense to me! Really if steps are optional there should probably be a consistent pattern for those anyways, so making the choice for italics here would have been confusing against other pages.

src/docs/framework-integration/angular.md
sean-perkins: The output targets generates an exported constant called `COMPONENTS` that is an array of all the component class declarations.

If you are creating a library, you would likely make use of this constant. Otherwise you would need to individually import and export all generated component classes (which is prone to errors when a developer deletes a deprecated component from Stencil, but forgets to update their Angular module). 

Open to suggestions here on how to best explain this is a generated constant value available to you. I'm not a huge fan of us calling the directory `stencil-generated` in our examples. I think including `stencil-` confuses the scope.

src/docs/framework-integration/angular.md
```diff
@@ -39,211 +43,179 @@ top-most-directory/
...
+### Adding Angular Output Target
 
-Now that the project structure is set up, we can install the Angular Output Target package. This package contains the Angular wrapper function that we will use to generate our Angular wrapped components. To install the Angular Output Target package, run the following command in your Stencil project directory
+Install the `@stencil/angular-output-target` dependency to your Stencil component library package.
 
-`npm i @stencil/angular-output-target`
-
-or
-
-`yarn add @stencil/angular-output-target`
```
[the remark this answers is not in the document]
sean-perkins: Re-added. Let me know if there's a more desired format for them. I know that Docusaurus has a toggle for auto-generating NPM vs. Yarn usage examples, but we aren't using Docusaurus for the Stencil docs.

src/docs/framework-integration/angular.md
```diff
@@ -39,211 +43,179 @@ top-most-directory/
...
 
-Within your Stencil’s config file, you can import the Angular wrapper function for use within the outputTargets array. If you copy and paste this, ensure you update the “my-workspace” and “my-lib” to coincide with what you named your component library, as well as to update the `componentCorePackage` to the name of your Stencil component library.
+```ts
+import { angularOutputTarget } from '@stencil/angular-output-target';
 
-```tsx
-import { angularOutputTarget as angular } from '@stencil/angular-output-target';
 export const config: Config = {
-  namespace: 'demo',
+  namespace: '',
```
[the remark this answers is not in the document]
sean-perkins: Updated to be `component-library` to match the other usage examples in the API documentation. I think we should likely sweep these and make them align with the default namespace in the future. Thoughts?

src/docs/framework-integration/angular.md
```diff
@@ -39,211 +43,179 @@ top-most-directory/
...
+If the build is successful, you will now have contents in the file specified in `directivesProxyFile`.
 
-Once the build is complete, you will see new files in your Angular component library's directory!
+You can now finally import and export the generated component wrappers for your component library. For example, in your main `AppModule` you may do the following:
 
+```ts
+import { COMPONENTS } from './stencil-generated/components';
 
-```
-top-most-directory/
```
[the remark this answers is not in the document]
sean-perkins: Is it necessary to show the whole folder-tree here to validate? The only contents that should change between running the command is if there is auto-generated contents in your proxy file path.

My goal was to narrow the context of what information you should be looking at, at this build step. We already show the directory tree in the project structure section.

Is there an alternative to this statement or additional context:
> If the build is successful, you will now have contents in the file specified in `directivesProxyFile`.

That would help with developer validation?

src/docs/framework-integration/angular.md
```diff
@@ -39,211 +43,179 @@ top-most-directory/
...
 
-Your Stencil library will be the location that you write your web components. 
+> If you already have a Stencil component library, skip this section.
 
 ```bash
 npm init stencil components stencil-library
 ```
 
-### Install the Angular Output Target in your Stencil Component Library
+### Adding Angular Output Target
```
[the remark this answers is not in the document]
sean-perkins: Good to be aware of! Thanks for the context. My concern with the existing title is that it creates line wrapping the side-navigation and is quite verbose.

src/docs/framework-integration/angular.md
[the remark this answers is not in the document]
rwaskiewicz: Agreed. I think the tricky part is that our Angular and React documentation tries to be both accommodating existing projects (e.g. we say things like "If you already have a Stencil component library, skip this section") as well as being tutorial-like for folks that don't.  In the case of the former, the namespace will be set already (likely to whatever the top level directory containing the stencil library is named), whereas the latter we have "stencil-library".  I don't think this problem is unsolvable by any means, just something we'll want to be aware of moving forward.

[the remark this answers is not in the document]
rwaskiewicz: Actually, we may want to call this 'stencil-library', because that namespace is populated (as 'stencil-library') by the command we provide in the previous section:
```
npm init stencil components stencil-library
```

src/docs/framework-integration/angular.md
rwaskiewicz: Hmm I think I'm missing something then. I can't find a `COMPONENTS` constant in my project.  I've run through generating a new stencil component and angular libraries, with the directions in this PR. Is there somewhere specific I should be looking?

Screen Shot 2022-03-01 at 1 29 00 PM


I looked in the output targets repo for 'COMPONENTS' (whole word, case sensitive) and couldn't find it either 🤔

sean-perkins: Great catch! I was getting my wires crossed with my output targets PR 🤦  we have an option that allows generating a constant called `DIRECTIVES`: https://github.com/ionic-team/stencil-ds-output-targets/blob/main/packages/angular-output-target/src/generate-angular-directives-file.ts#L28

Which is the incorrect label for the values. In my output targets refactor, I correct this name to align with the Angular community's naming convention.

I will need to update this PR to reference that variable name in the interim though.

src/docs/framework-integration/angular.md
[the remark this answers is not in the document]
rwaskiewicz: Yeah, that's fair - I think the single statement would suffice here 👍

rwaskiewicz (approved):

src/docs/framework-integration/angular.md
```diff
@@ -39,211 +43,199 @@ top-most-directory/
...
+export class ExampleLibraryModule {}
+```
 
-In order to make the generated files available to your Angular component library and it’s consumers, you’ll need to export everything from within your entry file - commonly the `public-api.ts` file. To do this, you’ll write:
+Any components that are included in the `exports` array should additionally be exported in your main entry point (either `public-api.ts` or `index.ts`). Skipping this step will lead to Angular Ivy errors when building for production.
 
-```tsx
-export * from './lib/stencil-generated/components';
+```ts
+export { DIRECTIVES } from './stencil-generated';
```
jstnjs, line 131: Should be `'./lib/stencil-generated'` right? But even then, for some reason it doesn't work. `export * from './lib/stencil-generated/components';` does!

jstnjs: Nevermind it does work with `'./lib/stencil-generated'`

PR #842, opened 2022-02-28
docs(support): add support policy + versioning docs

rwaskiewicz:
this pr adds 2 pages to our documentation:
1. a support policy
2. an explainer on how Stencil shall use semver

both documents are heavily borrowed from the Ionic
framework, where the former is borrowed from
https://ionicframework.com/docs/reference/support
and the latter from https://ionicframework.com/docs/reference/versioning

### DIFF
--- a/src/assets/docs-structure.json
+++ b/src/assets/docs-structure.json
@@ -430,7 +430,8 @@
     "children": [
       {
         "text": "Support Policy",
-        "filePath": "/assets/docs/reference/support-policy.json"
+        "filePath": "/assets/docs/reference/support-policy.json",
+        "url": "/docs/support-policy"
       }
     ]
   }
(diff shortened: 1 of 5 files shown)


src/assets/docs-structure.json
```diff
@@ -424,5 +424,20 @@
...
+        "url": "/docs/support-policy"
+      },
+      {
+        "text": "Versioning",
+        "filePath": "/assets/docs/reference/versioning.json",
+        "url": "/docs/versioning"
+      }
+    ]
   }
 ]
```
rwaskiewicz, line 443: This file is autogenerated, so I didn't bother adding a newline here

ltm (approved):

a-giuliano (approved):
Excellent

PR #845, opened 2022-03-08
docs(custom-elements): add generateTypeDeclarations doc

rwaskiewicz:
this commit adds documentation for the `generateTypeDeclarations` flag
on the `dist-custom-elements` output target

See https://github.com/ionic-team/stencil/pull/3270 for implementation details

### DIFF
--- a/src/docs/output-targets/custom-elements.md
+++ b/src/docs/output-targets/custom-elements.md
@@ -51,6 +51,20 @@ This flag defaults to `false` when omitted from a Stencil configuration file.
 
 > Note: At this time, components created not using JSX may not be automatically defined. This is a known limitation of the API and users should be aware of it
 
+### generateTypeDeclarations
+
+By default, type declaration files (`.d.ts` files) are only generated for the `dist-custom-elements` output target when
+the [`dist` output target](/docs/distribution) is also declared in a Stencil project's configuration. This behavior
+isn't always desirable, as not all users have need for files emitted by the `dist` output target. To generate type
+declaration files for the `dist-custom-elements`, the experimental `generateTypeDeclarations` field can be set to 
+`true`.
+
+Setting this flag to `true` will generate type declaration files for the `dist-custom-elements` output target. Type
+declaration files will be placed in the `dist/types` directory in the root of a Stencil project. At this time, this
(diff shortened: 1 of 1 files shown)


a-giuliano (approved):

src/docs/output-targets/custom-elements.md
```diff
@@ -51,6 +51,20 @@ This flag defaults to `false` when omitted from a Stencil configuration file.
 
 > Note: At this time, components created not using JSX may not be automatically defined. This is a known limitation of the API and users should be aware of it
 
+### generateTypeDeclarations
+
+By default, type declaration files (`.d.ts` files) are only generated for the `dist-custom-elements` output target when
+the [`dist` output target](/docs/distribution) is also declared in a Stencil project's configuration. This behavior
+isn't always desirable, as not all users have need for files emitted by the `dist` output target. To generate type
```
a-giuliano, line 58: nit
```suggestion
isn't always desirable, as not all users need the files emitted by the `dist` output target. To generate type
```

src/docs/output-targets/custom-elements.md
```diff
@@ -51,6 +51,20 @@ This flag defaults to `false` when omitted from a Stencil configuration file.
 
 > Note: At this time, components created not using JSX may not be automatically defined. This is a known limitation of the API and users should be aware of it
 
+### generateTypeDeclarations
+
+By default, type declaration files (`.d.ts` files) are only generated for the `dist-custom-elements` output target when
```
a-giuliano, line 56: Is it worth noting what the type declarations files are used for? Do we explain these somewhere else? If so, we should link here. If not, I'm not sure this is the right place to explain them for the first time, but it may be a worthwhile thing to explain somewhere else down the road.

rwaskiewicz: We _sort of_ do in other parts of the documentation, but agree we should add a formal suggestion (or reference to the TS docs) somewhere on the site. Created STENCIL-400 to track

a-giuliano: 👍

a-giuliano (approved):

PR #855, opened 2022-03-22
docs(assets): rewrite assets documentation

rwaskiewicz:
this commit includes a full rewrite of the existing assets
documentation. it renames the documentation from 'local assets' to
'assets', and sets up a permanent redirect for external links.

this commit introduces several nuances to using output targets other
than `dist` and `www`. because this is documentation, it does not fix
the rough patches in the core product itself. rather it seeks to
acknowledge them in order to raise awareness and gain signal around
places the core compiler can be improved


src/docs/guides/assets.md
```diff
@@ -0,0 +1,251 @@
+---
```
rwaskiewicz, line 1: FWIW, I did a `git mv local-assets.md assets.md`. Unsure what GitHub's heuristic for saying this file is new and `local-assets` has been removed is here

src/docs/guides/assets.md
```diff
@@ -0,0 +1,251 @@
...
+Oftentimes, this means there is only one asset base path per application using Stencil.
+
+## Resolution Overview
+
+The process of resolving an asset involves asking Stencil to build a path to the asset on the filesystem. 
+
+When an asset's path is built, the resolution is always done in a project's compiled output, not the directory containing the original source code.
+
+The example below uses the output of the [`www` output target](/docs/www) to demonstrate how assets are resolved.
+Although the example use the output of `www` builds, the general principle of how an asset is found holds for all output targets.
```
a-giuliano, line 42: typo
```suggestion
Although the example uses the output of `www` builds, the general principle of how an asset is found holds for all output targets.
```

src/docs/guides/assets.md
```diff
@@ -0,0 +1,251 @@
...
+
+To resolve the path to an asset, Stencil's [`getAssetPath()` API](/docs/assets#getassetpath) may be used.
+When using `getAssetPath`, the assets in the directory structure above are resolved relative to `build/`.
+
+The code sample below demonstrates the return value of `getAssetPath` for different `path` arguments.
+The return value is a path that Stencil has built to retrieve the asset on the filesystem.
+```ts
+// with an asset base path of "/build/":
+
+// '/build/assets/logo.png'
```
a-giuliano, line 68: Could we maybe put these comments inline with the code? It's a little confusing showing the result before the function call.

src/docs/guides/assets.md
```diff
@@ -0,0 +1,251 @@
...
+```
+src/
+└── components/
+    ├── assets/
+    │   ├── beach.jpg
+    │   └── sunset.jpg
+    └── stencil-asset.tsx
+```
+
+Below, the `stencil-asset` component will correctly load the assets based on it's `image` prop.
```
a-giuliano, line 101: Totally get what you're going for here, but could we call this component something different? Maybe just `my-component`? Though I think most will understand this, I fear that `stencil-asset` may suggest to some that the component is an asset itself.

src/docs/guides/assets.md
```diff
@@ -0,0 +1,251 @@
...
+ * @param path the path of the asset to build a URL to
+ * @returns the built URL
+ */
+declare function getAssetPath(path: string): string;
+```
+
+The code sample below demonstrates the return value of `getAssetPath` for different `path` arguments, when an asset base path of `/build/` has been set.
+```ts
+// with an asset base path of "/build/":
+// "/build/"
```
a-giuliano, line 215: Same comment as above about inlining

src/docs/guides/assets.md
```diff
@@ -0,0 +1,251 @@
...
+getAssetPath('other-assets/font.tiff'); 
+```
+
+## Making Assets Available
+
+In order to be able to find assets at runtime, they need to be found on the filesystem from the output of a Stencil build.
+In other words, we need to ensure they exist in the distribution directory.
+This section describes how to make assets available under the asset base path.
+
+### assetsDirs
```
a-giuliano, line 82: Here's what I'm thinking. This section is probably the most practically important part of the docs as it shows how to actually use assets. Being so important, I wonder if this should be higher up in the doc so it's more easily available to those who just want to see the implementation quickly. I think the order you have right now makes perfect logical sense, but might not be the most practical.

src/docs/guides/assets.md
rwaskiewicz: We could, but unfortunately that leads to creating a code block that requires horizontal scrolling on a smaller device (even on my 13" laptop).  LMK what you think, but this felt like the more readable of the two

src/docs/guides/assets.md
rwaskiewicz: I'm not sure I agree - `assetsDirs` doesn't work for `dist-custom-elements` output target, and I'm a little concerned we'd be highlighting something that doesn't work for one of our default output targets.  

I think this option is important for sure, but in Stencil's current implementation of assets, the quick implementation is not something folks should be jumping to (IMO).  Assets are something that Stencil can't do a ton of legwork for on a developer's behalf.  Something like this is going to be for developers to work out themselves. I think this will especially true as they explore architectures that include one or more layers of component libraries

src/docs/guides/assets.md
rwaskiewicz: 👍 Done in fb492c5d0..8c9b799d7

src/docs/guides/assets.md
a-giuliano: Ah I didn't consider that. I prefer your version then.

src/docs/guides/assets.md
a-giuliano: Well defended. And I don't feel particularly passionate about this. Good to leave as is.

src/docs/guides/assets.md
a-giuliano: 👍

a-giuliano (approved):

Issue #751, opened 2021-07-08
Docs: Update Vue integration to include @stencil-vue-output-target


bmpickford:
Page Ref: https://stenciljs.com/docs/vue

### Description
Currently the documentation describes a manual integration into Vue. An [output target for Vue](https://github.com/ionic-team/stencil-ds-output-targets/tree/master/packages/vue-output-target) now exists so this could be added to the documentation

### Possible Solution
Update site docs to match https://github.com/ionic-team/stencil-ds-output-targets#vue

rwaskiewicz:
This has been completed in #864 - thanks again!

PR #864, opened 2022-04-08
docs(vue-output-target): vue 3 output target walkthrough

sean-perkins:
Updates the documentation and walkthrough for the Vue Output Target for Vue 3. Consider these docs still a work in progress, happy to receive feedback/tweaks on the contents.

### DIFF
--- a/src/docs/framework-integration/vue.md
+++ b/src/docs/framework-integration/vue.md
@@ -8,50 +8,274 @@ contributors:
   - kensodemann
   - brysalazar12
   - iskanderbroere
+  - sean-perkins
 ---
 
 # Vue
 
-In order to use the custom element library within the Vue app, the application must be modified to define the custom elements and to inform the Vue compiler which elements to ignore during compilation. This can all be done within the `main.js` file.
+**Supports: Vue 3 • TypeScript 4.0+ • Stencil v2.9.0+**
 
-Assuming you’ve run `npm install --save test-components` beforehand, and that `test-component` is the name of our made up Web Components that we have published to npm, you import the components into the 'main.js' file by
+Stencil can generate Vue component wrappers for your web components. This allows your Stencil components to be used within a Vue 3 application. The benefits of using Stencil's component wrappers over the standard web components include:
 
-- importing the node module
-- telling Vue to ignore the custom element tags (see [docs](https://vuejs.org/v2/api/#ignoredElements))
-- binding the Stenciljs component code to the window object
+- Type checking with your components.
+- Integration with the router link and Vue router.
+- Optionally, form control web components can be used with `v-model`.
 
-```tsx
-import Vue from 'vue';
-import App from './App.vue';
+## Setup
 
-import { applyPolyfills, defineCustomElements } from 'test-components/loader';
+### Project Structure
 
-Vue.config.productionTip = false;
+We recommend using a monorepo structure for your component library with component wrappers. Your project workspace should contain your Stencil component library and the library for the generate Vue component wrappers.
 
-// Tell Vue to ignore all components defined in the test-components
-// package. The regex assumes all components names are prefixed
-// 'test'
-Vue.config.ignoredElements = [/test-\w*/];
+An example project set-up may look similar to:
 
-// Bind the custom elements to the window object
-applyPolyfills().then(() => {
-  defineCustomElements();
-});
-
-new Vue({
-  render: h => h(App)
-}).$mount('#app');
+```
+top-most-directory/
+└── packages/
+    ├── component-library/
+    │   └── src/
+    │       ├── lib/
+    │       └── public-api.ts
+    └── stencil-library/
+        ├── stencil.config.js
+        └── src/components
 ```
 
-The components should then be available in any of the Vue components
-```tsx
-render() {
-  return (
-    <div>
-      <test-stencil-component></test-stencil-component>
-    </div>
-  )
+#### Creating a Vue Component Library
+
+> If you already have a Vue component library, skip this section.
+
+The first time you want to create the component wrappers, you will need to have a Vue library package to write to.
+
+Using Lerna and Vue's CLI, generate a workspace and a library for your Vue component wrappers:
+
+```bash
+lerna init
+npm install typescript @types/node --save-dev
+# or if you are using yarn
+yarn add typescript @types/node --dev
+
+lerna create component-library
+# Follow the prompts and confirm
+cd packages/
+# Install Vue dependency
+npm install vue@3 --save-dev
+# or if you are using yarn
+yarn add vue@3 --dev
+# Add the stencil-library dependency
+lerna add stencil-library
+cd ..
+lerna bootstrap
+```
+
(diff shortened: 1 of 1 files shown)


src/docs/framework-integration/vue.md
```diff
@@ -8,50 +8,378 @@ contributors:
...
+```bash
+cd packages/
+npm init stencil components stencil-library
+cd stencil-library
+# Install dependencies
+npm install
+# of if you are using yarn
+yarn install
+cd ..
+lerna bootstrap
```
rwaskiewicz, line 57: We probably shouldn't assume any specific monorepo tooling. Folks using nx, bazel, etc. can probably figure out their equivalent command, but if someone's not coming from a monorepo tooling background and this is their first experience with it, they may be stuck scratching their heads here. My recommendation is we do a tooling-agnostic package linking here.

src/docs/framework-integration/vue.md
```diff
@@ -8,50 +8,378 @@ contributors:
...
+#### Creating a Vue Component Library
+
+> If you already have a Vue component library, skip this section.
+
+The first time you want to create the component wrappers, you will need to have a Vue library package to write to.
+
+Using Lerna and Vue's CLI, generate a workspace and a library for your Vue component wrappers:
+
+```bash
+lerna init
```
rwaskiewicz, line 69: Same comment regarding monorepo tooling. If I don't have lerna on my machine, so our instructions are "broken" for me on running `lerna init`.

src/docs/framework-integration/vue.md
```diff
@@ -8,50 +8,378 @@ contributors:
...
+npm install typescript @types/node --save-dev
```
rwaskiewicz, line 70: We may want to clarify that these should be run from the `top-most-directory/` - as a user I had to infer that from [this line](https://github.com/ionic-team/stencil-site/pull/864/files#diff-af403489927c5c8dffa93fe5aa0478c28408f6b160f7e3e029475f676a005321R76), which felt like it could have been a little smoother

src/docs/framework-integration/vue.md
rwaskiewicz: Otherwise, we should probably make a note that `lerna init` should be run on the `packages`(?) level before running this

src/docs/framework-integration/vue.md
```diff
@@ -8,50 +8,378 @@ contributors:
...
+  "scripts": {
+    "test": "echo \"Error: run tests from root\" && exit 1",
+    "build": "npm run tsc",
+    "tsc": "tsc -p ."
+  },
+  "publishConfig": {
+    "access": "public"
+  }
+}
+```
```
rwaskiewicz, line 148: Consider using `diff` as the markdown type to help illustrate what's new:
```suggestion
```diff
{
+  "types": "lib/index.d.ts",
  "scripts": {
-   "test": "echo \"Error: run tests from root\" && exit 1"
+   "test": "echo \"Error: run tests from root\" && exit 1",
+    "build": "npm run tsc",
+    "tsc": "tsc -p ."
  },
+  "publishConfig": {
+    "access": "public"
+  }
}
```
```

File: src/docs/framework-integration/vue.md [hunk 5]
```diff
@@ -8,50 +8,378 @@ contributors:
...
+You can now build your Stencil component library to generate the component wrappers.
+
+```bash
+# Build the library and wrappers (from packages/stencil-library)
+npm run build
+# or if you are using yarn
+yarn run build
+```
+
+If the build is successful, you will not have contents in the file specifies in `proxiesFile`.
```
rwaskiewicz, line 188:
```suggestion
If the build is successful, you will now have contents in the file specified in `proxiesFile`.
```

src/docs/framework-integration/vue.md
```diff
@@ -8,50 +8,378 @@ contributors:
...
+};
+```
+
+This plugin will be used by Vue applications as follows:
+
+```ts
+import { ComponentLibrary } from 'vue-library';
+
+createApp(App).use(ComponentLibrary).mount('#app');
+```
```
rwaskiewicz, line 215: Does it make sense to have this snippet here? We repeat it not too much longer in the "Consumer Usage" section [here](https://github.com/ionic-team/stencil-site/pull/864/files#diff-af403489927c5c8dffa93fe5aa0478c28408f6b160f7e3e029475f676a005321R229-R236)

src/docs/framework-integration/vue.md
```diff
@@ -8,50 +8,378 @@ contributors:
...
 
-Vue provides several different ways to install and use the framework in an application. The above technique for integrating a Stencil custom element library has been tested on a Vue application that was created using the `vue-cli` with ES2015 and WebPack as primary options. A similar technique should work if the application was generated using other options.
+Update your `package.json`, adding the following options to the existing config:
+
+```json
+{
+  "types": "lib/index.d.ts",
+  "scripts": {
+    "test": "echo \"Error: run tests from root\" && exit 1",
+    "build": "npm run tsc",
```
rwaskiewicz, line 141: I don't think we ever explicitly call this in the documentation - does this ever get called by lerna? I'm not sure which command in the documentation is invoking this if so

src/docs/framework-integration/vue.md
sean-perkins: TIL about the `diff` markdown type, that's super cool!

src/docs/framework-integration/vue.md
sean-perkins: You would run this command when building and publishing/distributing your Vue library. I suppose that could be a useful sub section so there isn't extra information in the doc that isn't being used anywhere

src/docs/framework-integration/vue.md
sean-perkins: I added a new section that talks about how the walkthrough uses Lerna, but any monorepo solution would work. I now cover how to install Lerna globally on your machine, so that commands will work. I also outline what alternative behavior you would call if using a different monorepo solution.

Using a monorepo vs. package linking is valuable here, as we use some of the project scaffolding that lerna has to create a Vue base project.

rwaskiewicz (changes requested):
All in all LGTM, just a few typos + copy edits

rwaskiewicz:
FWIW, I believe this would close https://github.com/ionic-team/stencil-site/issues/751

src/docs/framework-integration/vue.md
```diff
@@ -8,50 +8,410 @@ contributors:
...
+
+> If you already have a monorepo, skip this section.
+
+```bash
+# From your top-most-directory/
+lerna init
+# or if you are using other monorepo tools, initialize a workspace
+npm install typescript @types/node --save-dev
+# or if you are using yarn
+yarn add typescript @types/node --dev
```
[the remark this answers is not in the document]
rwaskiewicz: This isn't as funky as it was to me originally, dismissing this comment

rwaskiewicz (approved):

PR #869, opened 2022-04-29
docs(support): add current jest 28 support status

rwaskiewicz:
this commit updates the documentation site's support status page to
state that jest v28 is not supported at this time, but will be in the
future


alicewriteswrongs (approved):
lgtm!

PR #870, opened 2022-05-02
docs(dist): fix some confusing documentation for `dist-custom-elements`

alicewriteswrongs:
It looks like some documentation got copy-pasted from the
`dist-dist-custom-elements-bundle` documentation, but as far as I can
tell it does not apply to the `dist-custom-elements` output target.

This just edits things a bit to clarify 1) that the component files are
written to a directory (`dist/components`) and 2) that each file written
will export a single component.


rwaskiewicz (approved):
👍

PR #872, opened 2022-05-04
chore(docs): reorder guides table of contents

rwaskiewicz:
this commit reorganizes the table of contents for the guides section of
the doc site. unlike other sections, the documents in this section
aren't directly related to one another. as a result, there's no natural
'flow' between the sections. as a result, finding the guide you're
looking for requires a little extra effort to scan through the entire
contents until you find what you're looking for (today). by
alphabetizing this list, the hope is that it makes finding a specific
guide a little bit easier
Screen Shot 2022-05-04 at 5 06 10 PM


alicewriteswrongs (approved):
💯

alicewriteswrongs:
related q: should we remove the 'Deno' guide?

alicewriteswrongs:
yeah I think that makes sense to me, if we redirect to the support policy then the (I think probably small) number of users who might try to navigate to the old link, maybe linked from somewhere else or an outdated cache on a search engine will at least be able to ctrl-f "deno" and find the right info

rwaskiewicz:
👍 https://github.com/ionic-team/stencil-site/pull/874 removes the deno guide from the site

PR #879, opened 2022-05-19
docs(started): update starter text

rwaskiewicz:
update the `npm init stencil` text after releasing v3.0.0 of the
create-stencil cli (which moved the app starter to the community)

### DIFF
--- a/src/docs/introduction/getting-started.md
+++ b/src/docs/introduction/getting-started.md
@@ -23,7 +23,7 @@ updated Node before continuing.
 Stencil can be used to create standalone components, or entire apps. When running `npm init stencil`, you will be
 provided with a prompt so that you can choose the type of project to start.
 
-```bash
+```text
 ? Select a starter project.
 
 Starters marked as [community] are developed by the Stencil Community,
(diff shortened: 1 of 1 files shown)


src/docs/introduction/getting-started.md
```diff
@@ -23,11 +23,18 @@ updated Node before continuing.
...
 
-```bash
-? Pick a starter › - Use arrow-keys. Return to submit.
-
-❯  component     Collection of web components that can be used anywhere
-   app           Minimal starter for building a Stencil app or website
+```text
+? Select a starter project.
+
+Starters marked as [community] are developed by the Stencil
```
a-giuliano, line 29: What do you think of putting this after the choices? It feels more like a footnote to me

rwaskiewicz: This is meant to mirror what you get when you run `npm init stencil@latest`, and reflects the text that we see there:


Screen Shot 2022-05-17 at 9 13 03 AM

(https://github.com/ionic-team/create-stencil/pull/68 is the PR that implemented that verbiage)

We could do this in the doc site for sure, but we won't be able to propagate those changes back to the CLI as a limitation of the library that we're using to prompt folks for a starter type. LMK what you think which is better there

rwaskiewicz: (I did massage the text a _little_ to make it easier to read with the site's narrow center column though)

a-giuliano (approved):

PR #883, opened 2022-05-23
docs(dist-custom-elements): update documentation to reflect index.js changes

alicewriteswrongs:
This adds documentation of the change to the `dist-custom-elements`
output target introduced in ionic-team/stencil#3368. That PR adds
a re-export of all the built components from the `index.js` file,
which will support some existing workflows based on the
`dist-custom-elements-bundle` output target.

### DIFF
--- a/src/docs/output-targets/custom-elements.md
+++ b/src/docs/output-targets/custom-elements.md
@@ -37,6 +37,19 @@ import { HelloWorld } from 'my-library/dist/components/hello-world';
 customElements.define('hello-world', HelloWorld);
 ```
 
+The output directory will also contain an `index.js` file which imports and
+re-exports all of your components and their respective `defineCustomElement`
+helper functions, looking something like this:
+
+```tsx
+export { setAssetPath, setPlatformOptions } from '@stencil/core/internal/client';
+export { MyComponent, defineCustomElement as defineCustomElementMyComponent } from './my-component.js';
+export {
+  MyOtherComponent,
+  defineCustomElement as defineCustomElementMyOtherComponent
+} from './my-other-component.js';
+```
+
 ## Config
 
 ### autoDefineCustomElements
--- a/src/docs/output-targets/custom-elements.md
+++ b/src/docs/output-targets/custom-elements.md
@@ -37,9 +37,9 @@ import { HelloWorld } from 'my-library/dist/components/hello-world';
 customElements.define('hello-world', HelloWorld);
 ```
 
-The output directory will also contain an `index.js` file which imports and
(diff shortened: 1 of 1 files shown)


src/docs/output-targets/custom-elements.md
```diff
@@ -37,6 +37,19 @@ import { HelloWorld } from 'my-library/dist/components/hello-world';
...
+
+```tsx
+export { setAssetPath, setPlatformOptions } from '@stencil/core/internal/client';
+export { MyComponent, defineCustomElement as defineCustomElementMyComponent } from './my-component.js';
+export {
+  MyOtherComponent,
+  defineCustomElement as defineCustomElementMyOtherComponent
+} from './my-other-component.js';
+```
+
```
rwaskiewicz, line 52: Can we add a snippet of how this could be used as well?  Right now, we tell folks that we'll export this file for them, but we don't give them much instruction as to how to use it compared to the first option (and probably should make that callout that this is an either/or proposition)

rwaskiewicz (changes requested):

src/docs/output-targets/custom-elements.md
alicewriteswrongs: that makes sense, I'll add an explanation of the specific use-case of setting this ad `module` in `package.json`, do you think there's anything else we should call out?

alicewriteswrongs: just realized there's a bit of a wrinkle with this...

src/docs/output-targets/custom-elements.md
```diff
@@ -96,9 +112,15 @@ The configs below provide examples of how to do this automatically with popular
 
 ## Distributing Custom Elements
 
-Your component library can be easily distributed on NPM, similar to how [`@ionic/core`](https://www.npmjs.com/package/@ionic/core) does it. From there, consumers of your library can decide how to import your library into their project. For the `dist-custom-elements`, the default import location would be `my-library/dist/components`, but this can get further configured within the `package.json` file.
+Your component library can be easily distributed on NPM, just like
```
a-giuliano, line 115:
```suggestion
Your component library can be distributed on NPM, just like
```

nit: as a general rule of thumb I try to avoid words like "easily" or "just". "Easy" to someone may not necessarily be easy for someone else

a-giuliano (approved):

src/docs/output-targets/custom-elements.md
alicewriteswrongs: 👍

Issue #885, opened 2022-05-24
Missing "module" script path in Custom Elements output target


lodybo:
The documentation page for the custom elements bundle is missing an example for the package.json `"module"` property.

On [line 102](https://github.com/ionic-team/stencil-site/blob/master/src/docs/output-targets/custom-elements.md?plain=1#L102) you can see that there is an instruction to set the `"module"` property, but the actual example it points to is missing.

In the example below that, the one specifying `@stencil/core` to be a dependency, we can see the `module` property being set:

```tsx
{
  "module": "dist/components/index.js",
  "dependencies": {
    "@stencil/core": "latest"
  },
  ...
}
```

Is it assumable that this is the path meant to be added in the missing instruction?

rwaskiewicz:
@lodybo Yes, you're right. The example there is meant to cover both statements (about setting the `module` property and setting `@stencil/core` as a dependency). However, it doesn't read well/looks confusing as is today.

PR #886, opened 2022-05-24
docs(custom-elements): clarify distribution example

rwaskiewicz:
this commit moves a statement about adding `@stencil/core` as
a dependency in `package.json` down. otherwise, it looks like we're
missing an example

fixes: #885

### DIFF
--- a/src/docs/output-targets/custom-elements.md
+++ b/src/docs/output-targets/custom-elements.md
@@ -101,8 +101,6 @@ Your component library can be easily distributed on NPM, similar to how [`@ionic
 
 To make the custom elements index the entry module for a package, set the `module` property in `package.json` to:
 
-Also be sure to set `@stencil/core` as a dependency of the package.
-
 ```tsx
 {
   "module": "dist/components/index.js",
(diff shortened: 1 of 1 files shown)


src/docs/output-targets/custom-elements.md
```diff
@@ -115,6 +113,8 @@ Also be sure to set `@stencil/core` as a dependency of the package.
 
 Note: If you are distributing both the `dist` and `dist-custom-elements`, then it's up to you to choose which one of them should be available in the `module` entry.
```
alicewriteswrongs, line 114: unrelated to your change: I've noticed that we often put notes like this inside of a blockquote (`> Note: ...`) so they stand out a bit.

Would it make sense to do that while we're in here? If we wanted to, we might want to move line 116 above the note

alicewriteswrongs (commented):
change looks good! it did indeed look like we were missing the example.

I just had one possible additional change while we're in here

src/docs/output-targets/custom-elements.md
rwaskiewicz: 👍 Done in https://github.com/ionic-team/stencil-site/pull/886/commits/67bc5a8636270d91f012be5f0d7b949a0b6c7c03 

Screen Shot 2022-05-24 at 1 14 37 PM

alicewriteswrongs: beautiful!

alicewriteswrongs (approved):
lgtm!

PR #888, opened 2022-06-03
docs(docs-readme): update docs-readme description of alternative CLI flag

alicewriteswrongs:
The documentation is currently out of date, describing the use of a
`--docs-readme` CLI flag which Stencil does not support. However, the
`--docs` CLI flag does what the docs say that the `--docs-readme` flag
should do, so we can point the user towards that instead.

This was reported in ionic-team/stencil#3405

STENCIL-472

### DIFF
--- a/src/docs/output-targets/docs-readme.md
+++ b/src/docs/output-targets/docs-readme.md
@@ -27,11 +27,15 @@ export const config: Config = {
 };
 ```
 
-Another option would be to add the flag `--docs-readme`, such as:
+Another option would be to use the `--docs` CLI flag, like so:
 
 ```bash
-stencil build --docs-readme
+stencil build --docs
 ```
+
+This will cause the Stencil compiler to automatically add a `docs-readme`
+output target to your configuration, without having to save it in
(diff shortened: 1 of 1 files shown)


src/docs/output-targets/docs-readme.md
```diff
@@ -27,12 +27,16 @@ export const config: Config = {
...
 
 ```bash
-stencil build --docs-readme
+stencil build --docs
 ```
 
+This will cause the Stencil compiler to automatically add a `docs-readme`
+output target to your configuration, without having to save it in
+`stencil.config.ts`.
+
```
rwaskiewicz, line 39: This sentence may be a little confusing - it states it adds a docs-readme output target to your configuration, but could be a little clearer when it comes to this not being persisted in your `stencil.config.ts`.  What do you think about:

```suggestion
This will cause the Stencil compiler to perform a one-time generation of README
files using the [`docs-readme`](https://stenciljs.com/docs/docs-readme) output target.
```

rwaskiewicz (approved):

PR #889, opened 2022-06-08
Add info about the files property in package.json

karthikeyan-arumugam-1558:
Since we [don't recommend](https://stenciljs.com/docs/distribution) `dist` to be tracked by version control, but expose it in package.json, people like me who are new get confused at the first glance. So thought describing the `files` property will make it clear that though we don't add `dist` to version control, publish it to NPM.


rwaskiewicz (approved):
Thanks!

PR #890, opened 2022-06-15
docs(vue): update vue library bootstrapping

rwaskiewicz:
this commit adds an additional change to the bootstrapping process
of a user's vue library, to ensure that the `main` field in `package.json`
is set to `lib/index.js`.

the reason for this change is that when users run `lerna create PROJ-NAME`,
by default, the `main` field takes on the value of 'PROJ-NAME', which isn't the
entrypoint we want folks to be using (I don't think)


alicewriteswrongs (approved):
lgtm!

PR #899, opened 2022-08-01
Docs/v3 default flag values

tanner-reits:
- Updates docs around config `sourceMaps` flag to callout new V3 default as `true`
- Updates docs around `dist-custom-elements` to default `generateTypeDeclarations` to `true` in V3

### DIFF
--- a/src/docs/config/overview.md
+++ b/src/docs/config/overview.md
@@ -180,9 +180,9 @@ Will generate the following comment:
 
 ## sourceMap
 
-*default: `false`*
+*default: `true`*
 
-When set to `true`, sourcemaps will be generated for a project.
+When set to `false`, sourcemaps will not be generated for a project.
 
 ```tsx
 sourceMap: true
@@ -200,8 +200,6 @@ compared to the minified result produced when `sourceMap` is not enabled.
 
 Developers are responsible for determining whether or not they choose to serve sourcemaps in each environment their
 components are served and implementing their decision accordingly.
-
-When omitted or set to `false`, sourcemaps will not be generated.
 
 ## srcDir
 
--- a/src/docs/config/overview.md
+++ b/src/docs/config/overview.md
@@ -182,10 +182,11 @@ Will generate the following comment:
 
 *default: `true`*
 
-When set to `false`, sourcemaps will not be generated for a project.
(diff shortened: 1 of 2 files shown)


src/docs/config/overview.md
```diff
@@ -180,9 +180,9 @@ Will generate the following comment:
...
 ## sourceMap
 
-*default: `false`*
+*default: `true`*
 
-When set to `true`, sourcemaps will be generated for a project.
+When set to `false`, sourcemaps will not be generated for a project.
 
 ```tsx
 sourceMap: true
```
rwaskiewicz, line 188: In the first few lines, we say:
- this field defaults to true
- say what happens when it is set to false
- provide an example using it, where we set it to true

which feels a little less than cohesive to me.  I would we restore the following lines ([line 1](https://github.com/ionic-team/stencil-site/pull/899/files#diff-689c19470ae2ed5ac0b4ba2bb534871dd960e99ae1638ca7da6bab934a5dd022L185)) ([line 2](https://github.com/ionic-team/stencil-site/pull/899/files#diff-689c19470ae2ed5ac0b4ba2bb534871dd960e99ae1638ca7da6bab934a5dd022L204)) and maybe set the example to show a type rather than just 'true' (or the union of `true | false`). As a result, we have a flow that looks something like:

- this field defaults to true
- say what happens when it is set to true
- provide an example using it, where we set it to true

So maybe something like:

---

## sourceMap

*default: `true`*

When omitted set to `true`, sourcemaps will be generated for a project.

```tsx
sourceMap: true | false
```

https://github.com/ionic-team/stencil-site/pull/899/files#diff-689c19470ae2ed5ac0b4ba2bb534871dd960e99ae1638ca7da6bab934a5dd022L204 could even be hoisted up to coincide with the 'When set to `true`...' line: 

---
## sourceMap

*default: `true`*

When omitted or set to `true`, sourcemaps will be generated for a project.
When set to `false`, sourcemaps will not be generated.

```tsx
sourceMap: true | false
```

---

Thoughts?

rwaskiewicz (changes requested):

src/docs/config/overview.md
tanner-reits: Yep, I like that more!

rwaskiewicz (approved):

PR #903, opened 2022-08-11
feat: 🎸 add "--no-build"-documentation in CLI-docs

mariohamann:
This is tied to the following PR: https://github.com/ionic-team/stencil/pull/3516


rwaskiewicz (approved):
Thanks again!

mariohamann:
As @rwaskiewicz pointed out in https://github.com/ionic-team/stencil/pull/3516, there is already a similar flag available in Stencil – I updated my PR to that one!

PR #909, opened 2022-08-29
feat(output-targets): add section for `customElementsExportBehavior`

tanner-reits:
This commit modifies the docs for `dist-custom-elements` to add documentation for the `customElementsExportBehavior` config option

### DIFF
--- a/src/docs/output-targets/custom-elements.md
+++ b/src/docs/output-targets/custom-elements.md
@@ -38,11 +38,13 @@ customElements.define('hello-world', HelloWorld);
 ```
 
 The output directory will also contain an `index.js` file which exports some helper methods by default. The contents of the file
-will look something like (the contents may look different if [`customElementsExportBehavior`](#customelementsexportbehavior) is specified):
+will look something like:
 
 ```js
 export { setAssetPath, setPlatformOptions } from '@stencil/core/internal/client';
 ```
+
+> Note: The contents may look different if [`customElementsExportBehavior`](#customelementsexportbehavior) is specified!
 
 ## Config
 
@@ -52,10 +54,19 @@ This config option provides additional behaviors that will alter the default com
 for this target. The desired behavior can be set via the following in a project's Stencil config:
 
 ```ts
-{
-  type: 'dist-custom-elements',
-  customElementsExportBehavior: 'default' | 'single-export-module'
-}
+// stencil.config.ts
+import { Config } from '@stencil/core';
+
+export const config: Config = {
+  outputTargets: [
+    {
+      type: 'dist-custom-elements',
+      customElementsExportBehavior: 'default' | 'single-export-module',
+    },
+    // ...
+  ],
+  // ...
+};
 ```
 
 | Option                 | Description                                                                                                                                                                                                                                                                                                                                                                                               |
--- a/src/docs/output-targets/custom-elements.md
+++ b/src/docs/output-targets/custom-elements.md
@@ -98,6 +98,8 @@ Setting this flag to `false` will not generate type declaration files for the `d
 
 This flag defaults to `true` when omitted from a Stencil configuration file.
 
+> When set to generate type declarations, Stencil respects the export behavior selected via `customElementsExportBehavior` and generates type declarations specific to the content of that file.
+
 ## Making Assets Available
 
 For performance reasons, the generated bundle does not include [local assets](/docs/assets) built within the JavaScript output,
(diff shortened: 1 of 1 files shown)


src/docs/output-targets/custom-elements.md
```diff
@@ -37,24 +37,34 @@ import { HelloWorld } from 'my-library/dist/components/hello-world';
 customElements.define('hello-world', HelloWorld);
 ```
 
-The output directory will also contain an `index.js` file which exports all of
-your components and their respective `defineCustomElement` helper functions,
-looking something like this:
+The output directory will also contain an `index.js` file which exports some helper methods by default. The contents of the file
+will look something like (the contents may look different if [`customElementsExportBehavior`](#customelementsexportbehavior) is specified):
```
rwaskiewicz, line 41: Consider moving the content in the parenthesis to a separate sentence after this section. ATM it reads a little funky:

```
The contents of the file will look something like:

<CODE_EXAMPLE/>

Note, the contents may look different if [`customElementsExportBehavior`](#customelementsexportbehavior) is specified.

File: src/docs/output-targets/custom-elements.md [hunk 2]
```diff
@@ -37,24 +37,34 @@ import { HelloWorld } from 'my-library/dist/components/hello-world';
...
+
+This config option provides additional behaviors that will alter the default component export OR custom element definition behaviors
+for this target. The desired behavior can be set via the following in a project's Stencil config:
+
+```ts
+{
+  type: 'dist-custom-elements',
+  customElementsExportBehavior: 'default' | 'single-export-module'
+}
+```
```
rwaskiewicz, line 59: Consider adding some additional context to this example to make it a little clearer where this configuration belongs. Some Stencil configs can get rather large (e.g. Ionic Framework Core). Providing some context helps frame the configuration options a bit more:
```suggestion
```ts
// stencil.config.ts
import { Config } from '@stencil/core';

  outputTargets: [
    {
      type: 'dist-custom-elements',
      customElementsExportBehavior: 'default' | 'single-export-module'
    },
    // ...
  ],
// ...
}
```

src/docs/output-targets/custom-elements.md
```diff
@@ -37,24 +37,34 @@ import { HelloWorld } from 'my-library/dist/components/hello-world';
...
+| `default`              | No additional re-export or auto-definition behavior will be performed.<br><br>This value will be used if no explicit value is set in the config, or if a given value is not a valid option.                                                                                                                                                                                                               |
+| `single-export-module` | All component and custom element definition helper functions will be exported from the `index.js` file in the output directory (see [Defining Exported Custom Elements](#defining-exported-custom-elements) for more information on this file's purpose). This file can be used as the root module when distributing your component library, see [below](#distributing-custom-elements) for more details. |
```
rwaskiewicz, line 64: Should we add a comment about treeshaking here and that it won't always be possible to fully treeshake with this option?

rwaskiewicz (changes requested):

src/docs/output-targets/custom-elements.md
tanner-reits: It should always be possible once `autoDefineCustomElements` is migrated to be one of these options and not its own flag. I could add a comment for now with a "todo" to remove it later, but all that should happen before the v3 release anyway 🤷‍♂️

rwaskiewicz (approved):

src/docs/output-targets/custom-elements.md
```diff
@@ -66,8 +87,8 @@ will also automatically recursively define any child components as well.
 This flag defaults to `false` when omitted from a Stencil configuration file.
 
 > Note: At this time, components created not using JSX may not be automatically
-  defined. This is a known limitation of the API and users should be aware of
-  it
+> defined. This is a known limitation of the API and users should be aware of
+> it
 
 ### generateTypeDeclarations
```
alicewriteswrongs, line 93: should we have a note here about how this option's behavior is affected by the setting for `customElementsExportBehavior` ? I think it wouldn't need to be anything more complicated than "When set to generate type declarations, Stencil respects the export behavior selected via `customElementsExportBehavior` and generates type declarations specific to the content of that file" or something along those lines

alicewriteswrongs (changes requested):
just had one question

src/docs/output-targets/custom-elements.md
tanner-reits: I like that 👍

alicewriteswrongs (approved):
lgtm!

PR #910, opened 2022-08-29
docs(angular-output-target): update path for proxy output

sean-perkins:
Fixes the relative path for generating the proxy files.

The project directory structure is:
```
angular-workspace/
  projects/
    component-lib/src/lib/stencil-generated/
  stencil-library/
    stencil.config.ts
```

When generating the proxies (from the Stencil library) it only needs to move one directory up and into the `projects/` directory. Currently it is writing to:

```
angular-workspace/
  angular-workspace/
    projects/
      component-lib/src/lib/stencil-generated/
```

### DIFF
--- a/src/docs/framework-integration/angular.md
+++ b/src/docs/framework-integration/angular.md
@@ -91,8 +91,8 @@ export const config: Config = {
   outputTargets: [
     angularOutputTarget({
       componentCorePackage: 'your-stencil-library-package-name',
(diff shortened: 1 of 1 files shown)


sean-perkins:
Closing... technically this is correct based on where the `stencil-library` project is generated:
```
angular-workspace/
stencil-library/
```

Although you can generate the stencil library inside the angular workspace. I'll put some additional thought if there is a better way to make this clearer.

PR #911, opened 2022-08-30
docs(output-targets): docs for `transformAliasedImportPaths` on `dist`

tanner-reits:
This commit adds documentation to the `dist` output target docs for the `transformAliasedImportPathsInCollection` config option

### DIFF
--- a/src/docs/output-targets/dist.md
+++ b/src/docs/output-targets/dist.md
@@ -44,9 +44,15 @@ By default, before each build the `dir` directory will be emptied of all files. 
 
 This flag defaults to `true` when omitted from a Stencil configuration file.
 
+### collectionDir
+
+The `collectionDir` config specifies the output directory within the [distribution directory](#dir) where the transpiled output of Stencil components will be written.
+
+This option defaults to `collection` when omitted from a Stencil configuration file.
+
 ### transformAliasedImportPathsInCollection
 
-This option will allow [path aliases](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) defined in a project's `tsconfig.json` to be transformed into relative paths in the output code for the implicit `dist-collection` output target. This does not affect imports for external packages.
+This option will allow [path aliases](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) defined in a project's `tsconfig.json` to be transformed into relative paths in the code output under the [collectionDir](#collectiondir) subdirectory for this output target. This does not affect imports for external packages.
 
 An example of path transformation could look something like:
 
--- a/src/docs/output-targets/dist.md
+++ b/src/docs/output-targets/dist.md
@@ -32,38 +32,37 @@ Luckily, both builds can be generated at the same time, and shipped in the same 
 
 ## Config
 
-| Property | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         | Default |
-|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
-| `dir`    | The `dir` config specifies the public distribution directory. This directory is commonly the `dist` directory found within [npm packages](https://docs.npmjs.com/getting-started/packages). This directory is built and rebuilt directly from the source files. Additionally, since this is a build target, all files will be deleted and rebuilt after each build, so it's best to always copy source files into this directory. It's recommended that this directory not be committed to a repository. | `dist`  |
-| `empty`  | By default, before each build the `dir` directory will be emptied of all files. To prevent this directory from being emptied, change this value to `false`.                                                                                                                                                                                                                                                                                                                          | `true`  |
-
+| Property                                  | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | Default |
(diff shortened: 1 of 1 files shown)


src/docs/output-targets/dist.md
```diff
@@ -32,38 +32,37 @@ Luckily, both builds can be generated at the same time, and shipped in the same
...
+| `transformAliasedImportPathsInCollection` | This option will allow [path aliases](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) defined in a project's `tsconfig.json` to be transformed into relative paths in the output code for the implicit `dist-collection` output target (i.e. something like `import * as utils from '@my-utils-path';` will become `import * as utils from '../path/to/utils';`). This does not affect imports for external packages. Currently, aliased paths are not transformed in the emitted code after running through the TS compiler.<br><br>If using the `dist-collection` output target directly, the same result can be achieved using the `transformAliasedImportPaths` flag on the target's config. | `false` |
```
rwaskiewicz, line 39: No action required - this table isn't properly formatted, but wasn't to begin with.  We could reformat it like so:
```suggestion
| Property                                  | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | Default |
|-------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
| `dir`                                     | The `dir` config specifies the public distribution directory. This directory is commonly the `dist` directory found within [npm packages](https://docs.npmjs.com/getting-started/packages). This directory is built and rebuilt directly from the source files. Additionally, since this is a build target, all files will be deleted and rebuilt after each build, so it's best to always copy source files into this directory. It's recommended that this directory not be committed to a repository.                                                                                                                                                                                                                           | `dist`  |
| `empty`                                   | By default, before each build the `dir` directory will be emptied of all files. To prevent this directory from being emptied, change this value to `false`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | `true`  |
| `transformAliasedImportPathsInCollection` | This option will allow [path aliases](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) defined in a project's `tsconfig.json` to be transformed into relative paths in the output code for the implicit `dist-collection` output target (i.e. something like `import * as utils from '@my-utils-path';` will become `import * as utils from '../path/to/utils';`). This does not affect imports for external packages. Currently, aliased paths are not transformed in the emitted code after running through the TS compiler.<br><br>If using the `dist-collection` output target directly, the same result can be achieved using the `transformAliasedImportPaths` flag on the target's config. | `false` |
```
but like I said, it wasn't like this beforehand, and this comment contradicts my advice in the next comment - take a look at that first and we can go from there

rwaskiewicz, line 39: The description for `transformAliasedImportPathsInCollection` is probably a little _too_ much information for a table cell. In it, we describe:
- path aliases in TS
- mention the implicit `dist-collection` output target, which is opaque to most users (they don't know that it's implicitly added, which may be confusing)
- provide an example of the transform
- state what this doesn't affect
- describe what it means to use `dist-collection` directly

I would consider breaking some of this out into a `Config` sections, like https://stenciljs.com/docs/custom-elements#config.

rwaskiewicz (changes requested):

src/docs/output-targets/dist.md
tanner-reits: This is what I was saying yesterday with some targets having different ways of documenting their configs. I went ahead and moved everything for `dist` out of a table format and into the structure used for `dist-custom-elements`. Definitely feels a bit cleaner

src/docs/output-targets/dist.md
```diff
@@ -32,38 +32,61 @@ Luckily, both builds can be generated at the same time, and shipped in the same
...
+
+### empty
+
+By default, before each build the `dir` directory will be emptied of all files. To prevent this directory from being emptied, change this value to `false`.
+
+This flag defaults to `true` when omitted from a Stencil configuration file.
+
+### transformAliasedImportPathsInCollection
+
+This option will allow [path aliases](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) defined in a project's `tsconfig.json` to be transformed into relative paths in the output code for the implicit `dist-collection` output target. This does not affect imports for external packages.
```
rwaskiewicz, line 49: My only concern here is that unless you work for Ionic/hack on Stencil, "implicit `dist-collection` output target" is a little confusing. Maybe the better thing to do is to rephrase this in terms on the output of the output target, in a way that's more relatable to an end user?
```suggestion
This option will allow [path aliases](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) defined in a project's `tsconfig.json` to be transformed into relative paths in the code output under the `dist/collection` subdirectory for this output target. This does not affect imports for external packages.
```

rwaskiewicz (changes requested):

rwaskiewicz (approved):

src/docs/output-targets/dist.md
tanner-reits: Good thought. I made that change, but also added the `collectionDir` option to the docs and linked that with what you were saying.

rwaskiewicz:
LGTM! Thanks!

Once @alicewriteswrongs ✔️'s we'll land the feature commit and queue this up for Monday's release!

alicewriteswrongs (approved):
lgtm! 👍

PR #912, opened 2022-09-01
docs(angular): generating single component angular modules

sean-perkins:
Updates the documentation of the resulting refactoring and feature behind single component Angular module generation. 

Link to the modified doc page: https://stencil-site-git-docs-angular-scam-ionic1.vercel.app/docs/angular

### DIFF
--- a/src/docs/framework-integration/angular.md
+++ b/src/docs/framework-integration/angular.md
@@ -295,6 +295,8 @@ The relative or absolute path to the file that will contain all of the generated
 
 If `true`, Angular modules will be generated for each Stencil component included in the compilation process. When using this option, library maintainers should export the entire contents of the `proxyDeclarationFile` from the entry point of their library.
 
+Single component Angular modules (SCAM) are a convention in the Angular community of having a single component declared and exported on an Angular module. This allows Angular's treeshaking to eliminate unused components and modules from the bundle and leads to reduced coupling and bundle sizes.
+
 ### includeImportCustomElements
 
 If `true`, Angular components will import and define elements from the `dist-custom-elements` build, rather than `dist`.
--- a/src/docs/framework-integration/angular.md
+++ b/src/docs/framework-integration/angular.md
@@ -285,9 +285,11 @@ The title of the Stencil package where components are available for consumers. T
 import { IonApp } from '@ionic/core/components/ion-app.js';
 ```
 
-### proxyDeclarationFile
+### proxyDeclarationFile (previously `directivesProxyFile`)
 
 The relative or absolute path to the file that will contain all of the generated component wrapper definitions and optionally Angular modules during the compilation process. This file should be either exported to the consumers of your library or directly referenced in your entry point module of the library.
+
+> In `@stencil/angular-output-target@0.5.0` `directivesProxyFile` was replaced with `proxyDeclarationFile`.
 
 ### createSingleComponentAngularModules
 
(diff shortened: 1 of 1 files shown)


src/docs/framework-integration/angular.md
```diff
@@ -91,14 +91,14 @@ export const config: Config = {
...
       componentCorePackage: 'your-stencil-library-package-name',
-      directivesProxyFile: '../angular-workspace/projects/component-library/src/lib/stencil-generated/components.ts',
+      proxyDeclarationFile: '../angular-workspace/projects/component-library/src/lib/stencil-generated/components.ts',
       directivesArrayFile: '../angular-workspace/projects/component-library/src/lib/stencil-generated/index.ts',
     }),
   ],
 };
 ```
 
-> The `directivesProxyFile` is the relative path to the file that will be generated with all of the Angular component wrappers. You will replace the file path to match your project's structure and respective names. You can generate any file name instead of `components.ts`.
```
rwaskiewicz, line 101: I think it might be a good idea to call out the fact `directivesProxyFile` once existed, at least for the time being.  If someone were to upgrade their library, even though this is a breaking change, we don't have anything it terms of a migration guide to say "replace `directivesProxyFile` with `proxyDeclarationFile`

rwaskiewicz (changes requested):

src/docs/framework-integration/angular.md
sean-perkins: Updated: https://stencil-site-c5lfk7ysm-ionic1.vercel.app/docs/angular#proxydeclarationfile-previously-directivesproxyfile

Let me know if there's a more preferred way of representing that. 

We are in pre 1.0, so it's a wild world anyways.

rwaskiewicz:
My only vote here would be to keep the (previously …) out of the header,
only because that will affect permalinks in the long run since AFAIK vercel
won’t allow us to set up a redirect on page fragments (but it’s pre-1.0, so
🤷)

tanner-reits (commented):
I think the changes make sense. Only thing I'd maybe consider is who we're targeting. If the idea is that we only expect people who know about SCAM to find and use this, then it's fine. But, if we want to expose this idea to more people we could have a short summary on the concept and its benefits/use-case or at least link out to some post/article. I know this is targeted to a small group of users rn, but if we're gonna build and support it, might as well make it useful to others 🤷‍♂️

sean-perkins:
@tanner-reits [updated](https://github.com/ionic-team/stencil-site/pull/912/commits/33a1bdffcae413e44fe7dc2aa566e4495a20cd29) to include a note about the high level purpose/benefits of single component Angular modules. Since they are more of a community convention, Angular does not have any direct guides or documentation that can be referenced here.

tanner-reits:
> @tanner-reits [updated](https://github.com/ionic-team/stencil-site/pull/912/commits/33a1bdffcae413e44fe7dc2aa566e4495a20cd29) to include a note about the high level purpose/benefits of single component Angular modules. Since they are more of a community convention, Angular does not have any direct guides or documentation that can be referenced here.

Perfect 👍 I think even something small like that can encourage consumers to go look into the concept further

rwaskiewicz (approved):

sean-perkins:
Closing for now. We've updated the component wrappers to `outputType`. Will re-evaluate if doc changes are needed in a new PR.

PR #913, opened 2022-09-09
docs(reactive-data): update @watch documentation 

rwaskiewicz:
this commit updates the documentation for `@Watch`. it primarily focuses
on discussing how to get the decorator to fire for class members that
are objects/arrays in order to provide clarification for how stencil
compares old/new values (by reference).

minor copy edits throughout the page have been made to improve clarity

### DIFF
--- a/src/docs/components/reactive-data.md
+++ b/src/docs/components/reactive-data.md
@@ -4,78 +4,208 @@ description: Reactive Data, Handling arrays and objects
 url: /docs/reactive-data
 contributors:
   - jthoms1
+  - rwaskiewicz
 ---
 
 # Reactive Data
 
-Stencil components update when props or state on a component change. For performance and simplicity, Stencil only compares references for changes, and will not rerender when data inside of an array or object changes.
-
+Stencil components update when props or state on a component change.
 
 ## Rendering methods
 
-When a component updates because of a state change (props or state change), the [`render()`](templating-jsx) method is scheduled to run.
+When a props or state change on a component, the [`render()` method](templating-jsx) is scheduled to run.
 
+## The Watch Decorator (`@Watch()`)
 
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):
Good improvements! 👍

src/docs/components/reactive-data.md
```diff
@@ -4,78 +4,208 @@ description: Reactive Data, Handling arrays and objects
...
 
-To update array or object data, use the following techniques, which are fast-becoming a core part of the modern JavaScript  toolbox.
+### Updating Arrays
 
-### Updating arrays
+For arrays, the standard mutable array operations such as `push()` and `unshift()` won't trigger a component update.
+These functions will change the content of the array, but won't change the reference to the array itself.
 
-For arrays, the standard mutable array operations such as `push()` and `unshift()` won't trigger a component update. Instead, non-mutable array operators should be used as they return a copy of a new array. These include `map()` and `filter()`, and the ES6 spread operator syntax.
+In order to make changes to an array, non-mutable array operators should be used.
```
alicewriteswrongs, line 76: would it be potentially clearer to say something like "the prop or state member should be assigned to a new array, created using a method like `.map()` or `filter()`, the spread operator syntax, or simply a new array literal" ? I wouldn't want to potentially give the impression that just calling `.map` will be enough (the difference being that you need to assign the new array which is the return value of map).

alicewriteswrongs (changes requested):
just had one wording quesiton

src/docs/components/reactive-data.md
alicewriteswrongs: yup, I think it's good :)

alicewriteswrongs (approved):
lgtm!

PR #914, opened 2022-09-09
docs(props): add mutable:true examples for arrays, obj

rwaskiewicz:
this commit adds examples for props that are marked as `mutable: true` at declaration time. its point/focus is to demonstrate that the _reference_ can be mutated within stencil without causing a warning, but mutable operations such as `Array.prototype.push` will not trigger a render

### DIFF
--- a/src/docs/components/properties.md
+++ b/src/docs/components/properties.md
@@ -695,8 +695,10 @@ export class ToDoListItem {
 #### Mutable Arrays and Objects
 
 Stencil compares Props by reference in order to efficiently rerender components.
-Setting `mutable: true` from a Prop that is an object or array allows the _reference_ to the Prop to change inside the component and trigger a render.
+Setting `mutable: true` on a Prop that is an object or array allows the _reference_ to the Prop to change inside the component and trigger a render.
 It does not allow a mutable change to an existing object or array to trigger a render. 
+
+For example, to update an array Prop:
 
 ```tsx
 import { Component, Prop, h } from '@stencil/core';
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):
Super helpful addition!

src/docs/components/properties.md
```diff
@@ -692,6 +692,105 @@ export class ToDoListItem {
 }
 ```
 
+#### Mutable Arrays and Objects
+
+Stencil compares Props by reference in order to efficiently rerender components.
+Setting `mutable: true` on a Prop that is an object or array allows the _reference_ to the Prop to change inside the component and trigger a render.
+It does not allow a mutable change to an existing object or array to trigger a render.
```
alicewriteswrongs, line 699: Possible wording suggestion that may flow a bit more smoothly?

```suggestion
Setting `mutable: true` on a Prop that is an object or array will cause Stencil to trigger a re-render whenever the _reference_ to the Prop changes inside the component.
However, mutating an existing object in-place will not trigger a re-render, since the reference to that object will not change. A common
pattern for forcing a re-render in this situation is to create a new array or object instead of mutating the existing instance in place.
```

alicewriteswrongs: Also a general question here, I think using a compare-by-reference check w/ `==` or `===` and rendering only when there's a new object (thereby avoiding any deep equality checks or anything silly like that) is a pretty common pattern in JS, are there any good explanatory articles we could link out to that break this down well? I checked MDN and didn't find a great explanation of compare-by-reference vs compare-by-value. Maybe we can assume a Stencil reader already knows this?

src/docs/components/properties.md
```diff
@@ -692,6 +692,105 @@ export class ToDoListItem {
...
+```
+
+In the example above, updating the Prop in place using `this.contents.push('Stencil')` would have no effect.
+Stencil does not see the change to `this.contents`, since it looks at the _reference_ of the Prop, and sees that it has not changed.
+This is done for performance reasons.
+If Stencil had to walk every slot of the array to determine if it changed, it would incur a performance hit.
+Rather, it is considered better for performance and more idiomatic to re-assign the Prop (in the example above, we use the spread operator).
+
+The same holds for objects as well.
+Objects should be updated using the spread operator in order to create a new reference and trigger a render:
```
alicewriteswrongs, line 750:
```suggestion
Rather than mutating an existing object in-place, a new object should be created using the spread operator. This object will be different-by-reference and therefore will trigger a re-render:  

```

alicewriteswrongs (changes requested):
a few suggested, non-essential wording changes

src/docs/components/properties.md
alicewriteswrongs: cool, that all sounds fine to me, especially since I didn't find something that did a great job breaking it down with a little googling

rwaskiewicz: 👍 anything else you want to see here in that case? (PR is still marked as 'Changes Requested')

alicewriteswrongs (approved):
lgtm!

PR #915, opened 2022-09-12
docs(jsx): rewrite the section on "loops" to clarify `key` usage

alicewriteswrongs:
Rewrite this section to clarify a little bit what the `key` attr will do and why you'd want to use it. The stuff there already was odd so I rewrote it for the most part.

Part of addressing ionic-team/stencil#3253

### DIFF
--- a/src/docs/components/templating-and-jsx.md
+++ b/src/docs/components/templating-and-jsx.md
@@ -204,7 +204,7 @@ render() {
 }
 ```
 
-If this array of children is dynamic, i. e., if any nodes may be added,
+If this array of children is dynamic, i.e., if any nodes may be added,
 removed, or reordered, then it's a good idea to set a unique `key` attribute on
 each element like so:
 
--- a/src/docs/components/templating-and-jsx.md
+++ b/src/docs/components/templating-and-jsx.md
@@ -176,32 +176,37 @@ render(){
 }
 ```
 
-## Loops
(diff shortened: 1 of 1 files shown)


src/docs/components/templating-and-jsx.md
```diff
@@ -176,32 +176,37 @@ render(){
...
   )
 }
 ```
 
-Each step through the `map` function creates a new JSX sub tree and adds it to the array returned from `map`, which is then drawn in the JSX tree above it.
-
-If your list is dynamic, i. e., it's possible to change, add, remove or reorder items, you should assign a unique `key` to each element to give it a stable identity. This enables Stencil to reuse DOM elements for better performance. The best way to pick a key is to use a string that uniquely identifies that list item among its siblings (often your data will already have IDs).
-
-> Do not use the `map`-function's index variable as a key. It does not represent a stable identity of an item as it can change if the order of the list changed or if you added an item to the beginning of the list. As such it is not suitable as a `key`.
+If this array of children is dynamic, i. e., if any nodes may be added,
```
rwaskiewicz, line 207:
```suggestion
If this array of children is dynamic, i.e., if any nodes may be added,
```

rwaskiewicz (changes requested):
Small edit nit

rwaskiewicz (approved):

PR #916, opened 2022-09-15
docs(output-target): custom elements auto-define export behavior

tanner-reits:
This commit updates `autoDefineCustomElements` from being a standalone config flag to an option on `customElementsExportBehavior`


src/docs/output-targets/custom-elements.md
```diff
@@ -69,22 +69,11 @@ export const config: Config = {
...
+| `single-export-module`        | All component and custom element definition helper functions will be exported from the `index.js` file in the output directory (see [Defining Exported Custom Elements](#defining-exported-custom-elements) for more information on this file's purpose). This file can be used as the root module when distributing your component library, see [below](#distributing-custom-elements) for more details. |
+| `auto-define-custom-elements` | A component and its children will be automatically defined with the `CustomElementRegistry` when the component's module is imported.                                                                                                                                                                                                                                                                      |
```
rwaskiewicz, line 76: Does it make sense for us to order this alphabetically? I'm kind of on the fence here, since "default" is kind of a special value and I don't know if I'd prefer to differentiate it here by having it at the top. I suppose that doesn't prevent us from doing something like
- default
- auto-define-custom-elements
- single-export-module

At the same time, there's really nothing wrong with
- auto-define-custom-elements
- default
- single-export-module

either. Thoughts?

tanner-reits: I like keeping `default` at the top, but alphabetizing the others makes sense.

rwaskiewicz (changes requested):
Sorry missed this the first time around (and GH won't let me comment inline) - can you update the type definition above to include 'auto-define-custom-elements'?
```diff
- customElementsExportBehavior: 'default' | 'single-export-module',
+ customElementsExportBehavior: 'default' | 'auto-define-custom-elements' | 'single-export-module',
```

rwaskiewicz (approved):

alicewriteswrongs (approved):
looks great!

PR #924, opened 2022-10-13
docs(custom-elements): add externalRuntime config option 

tanner-reits:
Starting to add some callouts around the impacts `externalRuntime` has on the `dist-custom-elements` output target. There isn't a lot here, but I couldn't think of a good overarching description that ties these behaviors together to add more context.

### DIFF
--- a/src/docs/config/overview.md
+++ b/src/docs/config/overview.md
@@ -48,6 +48,9 @@ bundles: [
 ]
 ```
 
+## devServer
+
+Please see the [Dev-Server docs](/docs/dev-server).
 
 ## enableCache
 
@@ -59,6 +62,9 @@ Stencil will cache build results in order to speed up rebuilds. To disable this 
 enableCache: true
 ```
 
+## extras
+
+Please see the [Extras docs](/docs/config-extras).
 
 ## globalScript
 
@@ -99,14 +105,14 @@ globalStyle: 'src/global/app.css'
 
 Check out the [styling docs](/docs/styling#global-styles) of how to use global styles in your app.
 
-## invisiblePrehydration
+## hashedFileNameLength
 
-*default: `true`*
+*default: `8`*
 
-When true, `invisiblePrehydration` will use Ionic's opinion to visually hide components before they are hydrated by adding an automatically injected style tag to the document's head. Setting `invisiblePrehydration` to `false` will not inject the style tag into the head, allowing you to style your web components pre-hydration. Please note that setting `invisiblePrehydration` to false will cause everything to be visible when your page is loaded, causing more prominent Flash of Unstyled Content (FOUC), however you can style your web component's fallback content to your preference.
+When the `hashFileNames` config is set to `true`, and it is a production build, the `hashedFileNameLength` config is used to determine how many characters the file name's hash should be.
 
 ```tsx
-invisiblePrehydration: true
+hashedFileNameLength: 8
 ```
 
 ## hashFileNames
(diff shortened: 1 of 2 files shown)


src/docs/output-targets/custom-elements.md
```diff
@@ -68,6 +68,15 @@ This flag defaults to `false` when omitted from a Stencil configuration file.
 > Note: At this time, components created not using JSX may not be automatically
   defined. This is a known limitation of the API and users should be aware of
   it
+### externalRuntime
+
+When `true`, this flag results in the following behaviors:
+
+1. Minification will follow what is specified in the [Stencil config](docs/config#minifyJs).
+2. Filenames will not be hashed
+3. When resolving all `@stencil/core/*` imports as a part of the rollup process all of these resolved modules will be marked as external and therefore not included in the bundle.
```
tanner-reits, line 77: Open to input here on more context or a way to group all these behaviors into a good description for what this really does 🤷‍♂️ Behaviors just feel disconnected from each other

src/docs/output-targets/custom-elements.md
```diff
@@ -68,6 +68,15 @@ This flag defaults to `false` when omitted from a Stencil configuration file.
 > Note: At this time, components created not using JSX may not be automatically
   defined. This is a known limitation of the API and users should be aware of
   it
+### externalRuntime
```
rwaskiewicz, line 71: Taking a step back, I took a look at the diff of using `externalRuntime` in a default Stencil project (`npm init stencil component`) this morning where I built the project twice with the following Stencil config output target settings:
```
    {
      type: 'dist-custom-elements',
      externalRuntime: false,
    },
    {
      type: 'dist-custom-elements-bundle',
      externalRuntime: false,
    },
```
```
    {
      type: 'dist-custom-elements',
      externalRuntime: true,
    },
    {
      type: 'dist-custom-elements-bundle',
      externalRuntime: true,
    },
```

It seems like this flag really only does anything for CEB. Is that right? Should we re-evaluate if this flag should exist for the CE output target?

tanner-reits: @rwaskiewicz I went and double-checked this and was seeing differences in `dist-custom-elements` with the flag on and off. Most notably is the bundling where, when `true`, the Stencil core code is not included in the bundle

tanner-reits: @rwaskiewicz I didn't run a diff, I was just looking at the output code when running with `externalRuntime: true` and `externalRuntime: false`. Were you not talking about the output? Maybe we can sync on this tomorrow morning?

src/docs/config/overview.md
```diff
@@ -119,17 +125,27 @@ During production builds, the content of each generated file is hashed to repres
...
 
+## invisiblePrehydration
 
-## hashedFileNameLength
-
-*default: `8`*
+*default: `true`*
 
-When the `hashFileNames` config is set to `true`, and it is a production build, the `hashedFileNameLength` config is used to determine how many characters the file name's hash should be.
+When true, `invisiblePrehydration` will use Ionic's opinion to visually hide components before they are hydrated by adding an automatically injected style tag to the document's head. Setting `invisiblePrehydration` to `false` will not inject the style tag into the head, allowing you to style your web components pre-hydration. Please note that setting `invisiblePrehydration` to false will cause everything to be visible when your page is loaded, causing more prominent Flash of Unstyled Content (FOUC), however you can style your web component's fallback content to your preference.
```
alicewriteswrongs, line 132: idk if we want to just leave this alone since it was just moved, but "Ionic's opinion" here strikes me as a really strange phrase - I would have no idea what that means if I came across it I think

src/docs/output-targets/custom-elements.md
alicewriteswrongs: Possibly it would be good to rephrase this along the same lines as the other entries in this list:

```suggestion
When `true`, this flag results in the following behaviors:

1. Minification will follow what is specified in the [Stencil config](docs/config#minifyJs).
2. Filenames will not be hashed
3. All imports from packages under `@stencil/core/*` will be marked as external and therefore not included in the bundle by Rollup.
```

alicewriteswrongs (changes requested):
just one suggestion and one question

src/docs/config/overview.md
tanner-reits: Agree this is super goofy. I'll go ahead and rephrase this slightly to get rid of that phrase while I'm in here

alicewriteswrongs (approved):
lgtm!

rwaskiewicz (approved):
Can you do me a favor and tweak the GH title/commit header before merging so this follows semantic commits?

PR #935, opened 2022-11-28
Migration docs

mhartington:


alicewriteswrongs (approved):
lgtm!

src/docs/introduction/upgrading-to-stencil-three.md
```diff
@@ -0,0 +1,295 @@
...
+for this target. The desired behavior can be set via the following in a project's Stencil config:
+
+```ts
+// stencil.config.ts
+import { Config } from '@stencil/core';
+export const config: Config = {
+  outputTargets: [
+    {
+      type: 'dist-custom-elements',
+      customElementsExportBehavior: 'default' | 'single-export-module',
```
tanner-reits, line 188: We're missing some options here:
```suggestion
      customElementsExportBehavior: 'default' | 'auto-define-custom-elements' | 'bundle' | 'single-export-module',
```

src/docs/introduction/upgrading-to-stencil-three.md
```diff
@@ -0,0 +1,295 @@
...
+| `default`              | No additional re-export or auto-definition behavior will be performed.<br><br>This value will be used if no explicit value is set in the config, or if a given value is not a valid option.                                                                                                                                                                                                                                                                                 |
+| `single-export-module` | All component and custom element definition helper functions will be exported from the `index.js` file in the output directory (see [Defining Exported Custom Elements](/docs/custom-elements#defining-exported-custom-elements) for more information on this file's purpose). This file can be used as the root module when distributing your component library, see [Distributing Custom Elements](/docs/custom-elements#distributing-custom-elements) for more details. |
```
tanner-reits, line 199: We'll need to update this to account for the missing options too:
```suggestion
| `default`                     | No additional re-export or auto-definition behavior will be performed.<br><br>This value will be used if no explicit value is set in the config, or if a given value is not a valid option.                                                                                                                                                                                                               |
| `auto-define-custom-elements` | A component and its children will be automatically defined with the `CustomElementRegistry` when the component's module is imported.                                                                                                                                                                                                                                                                      |
| `bundle`                      | A utility `defineCustomElements()` function is exported from the `index.js` file of the output directory. This function can be used to quickly define all Stencil components in a project on the custom elements registry.                                                                                                                                                                                 |
| `single-export-module`        | All component and custom element definition helper functions will be exported from the `index.js` file in the output directory (see [Defining Exported Custom Elements](#defining-exported-custom-elements) for more information on this file's purpose). This file can be used as the root module when distributing your component library, see [below](#distributing-custom-elements) for more details. |
```

tanner-reits: Or were you omitting those options because they're covered separately below?

tanner-reits (commented):
Not sure if what I commented on was intentional or not. If so, then I think we're g2g

src/docs/introduction/upgrading-to-stencil-three.md
rwaskiewicz: Ah, good catch. That's on me, grabbed the docs from an earlier commit

src/docs/introduction/upgrading-to-stencil-three.md
rwaskiewicz: Nope, better they all be here for now

rwaskiewicz:
@tanner-reits good catch! Applied your suggestions

tanner-reits (approved):

PR #941, opened 2022-12-05
docs(v3-upgrade): update npm tag for v3 prereleases

rwaskiewicz:
this commit updates the upgrade guide for the v2 version of the site. specifically, it changes the `npm install` scripts in the 'Getting Started' section to use teh `v3-next` tag, rather than a semver compliant version (`3`). we do this as v3 has not been officially released, making the script useless.

instead, we use an intentionally vague tag number that allows the dev team to update v3-next behind the scenes (from v3.0.0-alpha.0 to v3.0.0-alpha.1, etc.) without having to change the documentation for every release


tanner-reits (approved):

alicewriteswrongs (approved):
lgtm!

PR #945, opened 2022-12-09
feat(docs): removal of inlineDynamicImports to migration guide

tanner-reits:
This commit adds a section to the v3 migration guide for the removal of `inlineDynamicImports` from `dist-custom-elements`

Related to [Stencil PR #3897](https://github.com/ionic-team/stencil/pull/3897)

### DIFF
--- a/src/docs/introduction/upgrading-to-stencil-three.md
+++ b/src/docs/introduction/upgrading-to-stencil-three.md
@@ -224,6 +224,11 @@ export const config: Config = {
 };
 ```
 
+#### Remove `inlineDynamicImports` Configuration
+The `inlineDynamicImports` configuration option on `dist-custom-elements` has been removed. Previously, this option would throw an error at build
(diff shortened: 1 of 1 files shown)


src/docs/introduction/upgrading-to-stencil-three.md
```diff
@@ -224,6 +224,11 @@ export const config: Config = {
 };
 ```
 
+#### Remove `inlineDynamicImports` Configuration
+The `inlineDynamicImports` configuration option on `dist-custom-elements` has been removed. Previously, this option would throw an error at build
+time during the Rollup bundling process if the build contained multiple "inputs" (components). Based on the use cases for Stencil components libraries
```
rwaskiewicz, line 229: I don't think we need this last sentence, "Based on the use cases...", I don't think it adds much value to the reader. If they're curious as to why we removed this (beyond it being broken),  we can always point them to the PR where we rm'ed it

rwaskiewicz (approved):

alicewriteswrongs (approved):

PR #947, opened 2022-12-13
Docs/csp guide

tanner-reits:
Adds a guide to the "Guides" section for using CSP nonces in Stencil. This covers some basic introduction to what a CSP/nonce is as well as the niche behavior for ingesting nonces into Stencil.


rwaskiewicz:
@tanner-reits Yeah, if you can run `npm run build`, [which runs the `docs` script](https://github.com/ionic-team/stencil-site/blob/0a793ec1850962d3714baa2839413f1ed9b6589a/package.json#L7) and generates everything for you (or should 🤔)

tanner-reits:
@rwaskiewicz Ah thanks! That was the secret sauce, but I also copied the header/metadata section from the only doc in that directory that didn't have the url attribute set which was also causing problems 🙄 All should be well now with the Vercel preview!

src/docs/guides/csp-nonce.md
```diff
@@ -0,0 +1,117 @@
...
+---
+
+# Using Content Security Policy Nonces in Stencil
+
+[Content Security Policies (CSPs)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) can help protect an application from Cross-Site Scripting (XSS)
+attacks by adding a security layer to help prevent unauthorized code from running in the browser.
+
+In the world of Stencil applications, if your app leverages a restrictive CSP (i.e. anything other than 'unsafe-inline') and components without a Shadow
+DOM, you'll likely run into an error in the browser console stating the certain styles or scripts were not able to run because they violate the
+effective CSP.
```
rwaskiewicz, line 16:
```suggestion
An application that is served with a restrictive CSP other than `'unsafe-inline'` & contains web components without a Shadow DOM will likely run into errors on load. This is often first detected in the browser's console, which reports an error stating that certain styles or scripts violate the effective CSP.
```

src/docs/guides/csp-nonce.md
```diff
@@ -0,0 +1,117 @@
...
+
+This is an example of consuming the `dist` output in an Angular app's entrypoint:
+
+```ts
+// main.ts
+
+import { defineCustomElements, setNonce } from 'my-lib/loader';
+
+// Will set the `nonce` attribute for all scripts/style tags
+// i.e will run styleTag.setAttribute('nonce', 'r4nd0m')
```
rwaskiewicz, line 57:
```suggestion
// i.e. will run styleTag.setAttribute('nonce', 'r4nd0m')
```

src/docs/guides/csp-nonce.md
```diff
@@ -0,0 +1,117 @@
...
+platformBrowserDynamic()
+  .bootstrapModule(AppModule)
+  .catch(err => console.log(err));
+```
+
+#### WWW
+
+Unfortunately, setting `nonce` attributes gets a bit trickier when it comes to [SSR and SSG](docs/static-site-generation). As a `nonce` needs
+to be unique per page view, we cannot define/set the nonce at build time. So, this responsibility now falls on the
+[hydrate app](docs/hydrate-app)'s execution of runtime code.
```
rwaskiewicz, line 100: IIRC, these links need a leading slash. ATM, this points to docs/docs/hydrate-app in the generated docs

```suggestion
[hydrate app](/docs/hydrate-app)'s execution of runtime code.
```

src/docs/guides/csp-nonce.md
```diff
@@ -0,0 +1,117 @@
...
+
+##### SSR
```
rwaskiewicz, line 102: I don't think the Stencil Site supports 5 pound signs at the moment:
Screenshot 2022-12-15 at 3 37 26 PM

Can we reorder/restructure this doc to work around this limitation?

src/docs/guides/csp-nonce.md
```diff
@@ -0,0 +1,117 @@
...
+
+Since there is not an easy way (or any way) of exposing and executing helper functions to manipulate the outcome of the runtime code, Stencil
+has fallback behavior for pulling the `nonce` off of the window object.
+
+So, for SSR, your app can simply set the desired `nonce` value on `window.nonce` at bootstrap. Yes, this does involve some manual configuration
+for the code served by your server.
```
rwaskiewicz, line 108: Can we update this to clarify that 'bootstrap' should be when the page is rendered? I'd like to make it clear that this shouldn't just happen once, but every time a server serves the page.

rwaskiewicz (changes requested):

src/docs/guides/csp-nonce.md
```diff
@@ -79,11 +78,11 @@ This is an example of consuming the `dist-custom-elements` output in an Angular
 // main.ts
 
 import { defineCustomElements, setNonce } from 'my-lib/dist/components';
-// We'll assume we have `autoDefineCustomElements` enabled
+// Assume `autoDefineCustomElements` is enabled enabled
```
rwaskiewicz, line 81: How do we want to handle this once `autoDefineCustomElements` goes away in v3.0.0?

rwaskiewicz (commented):
Looks good, just one outstanding question for v3

src/docs/guides/csp-nonce.md
tanner-reits: Once this lands and we rebase v3, I can put up a small PR to change this comment to the export behavior instead

rwaskiewicz (approved):

src/docs/guides/csp-nonce.md
alicewriteswrongs: does this still need to get updated?

tanner-reits: Nope, sorry, forgot to mark this resolved. I replaced those headers with bolded elements. Maybe we can swap this back if the new docs site supports up to h5/h6

alicewriteswrongs (approved):
lgtm!

PR #948, opened 2022-12-13
docs(v3-upgrade): add note about deprecating fields in extras

alicewriteswrongs:
first commit: This starts a new section in the v3 upgrade docs for the fields in `config.extras` which will be marked as deprecated in Stencil v3. This first commit includes notes about the `dynamicImportShim` field. See https://github.com/ionic-team/stencil/pull/3895

PR #950, opened 2022-12-13
docs(v3-upgrade): add note about deprecating `cssVarsShim`

alicewriteswrongs:
This adds a note to the v3 upgrade guide about deprecating `extras.cssVarShim`.

**NOTE**: this PR has https://github.com/ionic-team/stencil-site/pull/948 as its base.

### DIFF
--- a/src/docs/introduction/upgrading-to-stencil-three.md
+++ b/src/docs/introduction/upgrading-to-stencil-three.md
@@ -106,6 +106,24 @@ export const config: Config = {
 };
 ```
 
+#### `cssVarShim`
+
+`extras.cssVarShim` causes Stencil to include a polyfill for [CSS
+variables](https://developer.mozilla.org/en-US/docs/Web/CSS/--*). For Stencil
+v3.0.0 this field is renamed to `__deprecated__cssVarsShim`, and thus to retain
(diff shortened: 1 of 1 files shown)


src/docs/introduction/upgrading-to-stencil-three.md
```diff
@@ -87,6 +87,24 @@ export const config: Config = {
 };
 ```
 
+#### `cssVarShim`
+
+`extras.cssVarShim` causes Stencil to include a polyfill for [CSS
+variables](https://developer.mozilla.org/en-US/docs/Web/CSS/--*). For Stencil
+v3.0.0 this field is renamed to `__deprecated__cssVarsShim`, and thus to retain
+the previous behavior the new option can be set in your projects
+`stencil.config.ts`:
```
tanner-reits, line 96: Part of this sounds a little weird to me. I think we should just copy-paste the same sentence used for `dynamicImportShim`. I don't see why we need different wording for essentially the same thing 🤷‍♂️

tanner-reits (changes requested):

alicewriteswrongs:
Agreed, just pushed that change

tanner-reits (approved):

PR #951, opened 2022-12-14
docs(angular-integration): revamp into complete guide

tanner-reits:
Re-works sections of our framework wrapper documentation to account for discovery during https://ionic-cloud.atlassian.net/browse/STENCIL-605


src/docs/framework-integration/angular.md
```diff
@@ -303,6 +317,92 @@ export const config: Config = {
 };
 ```
 
-### excludeComponents
+## FAQs
+
+### Do I have to use the `dist` output target?
```
tanner-reits, line 322: This section in the only new content for the FAQs. The rest was just moved to fall below the API section

tanner-reits:
@alicewriteswrongs Forgot to mention this earlier, buy Ryan called me out. The first two commits here are just formatting/restructure. The actual content changes are in [da9de97](https://github.com/ionic-team/stencil-site/pull/951/commits/da9de97f5169de9dc30ba41977c9ec2709aff5da) so you can view those without all the other diff chaos

src/docs/framework-integration/angular.md
```diff
@@ -204,38 +211,84 @@ Developers can now directly leverage your components in their template and take
...
-The title of the Stencil package where components are available for consumers. This is used during compilation to write the correct imports
-for components e.g.
+**Required**
+
+**Type: `string`**
+
+The title of the Stencil package where components are available for consumers (i.e. the `name` property value in your Stencil project's `package.json`).
+This is used during compilation to write the correct imports for components.
+
+For the Ionic Framework, a generated import would look like:
```
rwaskiewicz, line 223: I like the idea of an example here, but  I think this one assumes some knowledge of how Framework is structured (it assumes some knowledge of `core`, where to find it's directory in the repo and the `package.json` file in that directory).

It may be more beneficial to show a snippet from `npm init stencil component my-component-lib`'s `package.json` (the part that includes the name) and how _that_ name gets translated into an import statement.

src/docs/framework-integration/angular.md
```diff
@@ -194,88 +209,87 @@ Developers can now directly leverage your components in their template and take
...
 ### valueAccessorConfigs
 
-This lets you define which components should be integrated with ngModel (I.e. form components). It lets you set what the target prop is (I.e. `value`), which event will cause the target prop to change, and more.
+**Optional**
+
+**Default: `[]`**
+
+**Type: `ValueAccessorConfig[]`**
+
+This lets you define which components should be integrated with ngModel (I.e. form components). It lets you set what the target prop is (i.e. `value`),
```
rwaskiewicz, line 291:
```suggestion
This lets you define which components should be integrated with `ngModel` (i.e. form components). It lets you set what the target prop is (i.e. `value`),
```

rwaskiewicz (changes requested):

rwaskiewicz:
@tanner-reits Changes look good 👍 Is the idea going to add on to this PR for React/Vue, or separate PRs for each?

tanner-reits:
@rwaskiewicz Was planning on adding to this PR, but could also branch off to do separate PRs if you'd rather go that route. Only thing that really needs to happen I think is for the changes to each to land at the same time

rwaskiewicz:
@tanner-reits Either is fine with me - I didn't know whether to approve this as-is to get it ready to merge or to wait for the other docs to get updated (now that Angular is in a good spot)

rwaskiewicz (changes requested):
Had a few more comments after running through this again alongside the React Docs revamp

src/docs/framework-integration/angular.md
```diff
@@ -125,15 +164,19 @@ import { DIRECTIVES } from './stencil-generated';
 export class ExampleLibraryModule {}
 ```
 
-Any components that are included in the `exports` array should additionally be exported in your main entry point (either `public-api.ts` or `index.ts`). Skipping this step will lead to Angular Ivy errors when building for production.
+Any components that are included in the `exports` array should additionally be exported in your main entry point (either `public-api.ts` or
```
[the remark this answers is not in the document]
alicewriteswrongs: I'd second this, also when I ran through this tutorial I ended up with the following directory structure on the angular side:

```
.
├── README.md
├── angular.json
├── package-lock.json
├── package.json
├── projects
│   └── component-library
│       ├── README.md
│       ├── ng-package.json
│       ├── package.json
│       ├── src
│       │   ├── lib
│       │   │   ├── component-library.component.spec.ts
│       │   │   ├── component-library.component.ts
│       │   │   ├── component-library.module.ts
│       │   │   ├── component-library.service.spec.ts
│       │   │   ├── component-library.service.ts
│       │   │   └── stencil-generated
│       │   │       ├── angular-component-lib
│       │   │       │   └── utils.ts
│       │   │       ├── components.ts
│       │   │       └── index.ts
│       │   └── public-api.ts
│       ├── tsconfig.lib.json
│       ├── tsconfig.lib.prod.json
│       └── tsconfig.spec.json
└── tsconfig.json

6 directories, 20 files
```

that's in the `angular-workspace` directory. So I have a `public-api.ts` file in `projects/component-library/src`, but to get to the Stencil stuff the relative path shown below isn't right, it should be `./lib/stencil-generated`, not `./stencil-generated`.

Maybe the `public-api.ts` should be in `src/lib` though?

src/docs/framework-integration/angular.md
```diff
@@ -157,9 +200,11 @@ yarn link name-of-your-stencil-package
 
 The name of your Stencil package should match the `name` property from the Stencil component library's `package.json`.
 
-Your component libraries are now linked together. You can make changes in the Stencil component library and run `npm run build` to propagate the changes to the Angular component library.
+Your component libraries are now linked together. You can make changes in the Stencil component library and run `npm run build` to propagate the
+changes to the Angular component library.
 
-> **NOTE**: As an alternative to `npm link` , you can also run `npm install` with a relative path to your Stencil component library. This strategy, however, will modify your `package.json` so it is important to make sure you do not commit those changes.
+> **NOTE**: As an alternative to `npm link` , you can also run `npm install` with a relative path to your Stencil component library. This strategy,
+> however, will modify your `package.json` so it is important to make sure you do not commit those changes.
```
alicewriteswrongs, line 207: maybe this is very obvious if you're an angular-er, but it might be nice to include a final step which just includes whatever commands are necessary to get the angular dev server running with the stencil components built and included, so someone following this tutorial can assure themselves that they've created something that works correctly

src/docs/framework-integration/angular.md
```diff
@@ -98,9 +133,13 @@ export const config: Config = {
 };
 ```
 
-> The `directivesProxyFile` is the relative path to the file that will be generated with all of the Angular component wrappers. You will replace the file path to match your project's structure and respective names. You can generate any file name instead of `components.ts`.
+> The `directivesProxyFile` is the relative path to the file that will be generated with all of the Angular component wrappers. You will replace the
```
alicewriteswrongs, line 136: it might be good to call out above that `componentCorePackage` has to match whatever name you gave the stencil component library.

If the user has entered the commands exactly as in the tutorial this will be `"stencil-library"`, so maybe we could change the value to that so it 'just works' ?

src/docs/framework-integration/angular.md
```diff
@@ -125,15 +164,19 @@ import { DIRECTIVES } from './stencil-generated';
...
 ```
 
-### Link your packages (optional)
+> **NOTE:** The default behavior does not automatically define the Web Components needed by the proxy components. You can manually define the Web
+> Components following the [documentation for the dist output target](/docs/custom-elements#defining-exported-custom-elements).
+
+### Link Your Packages (Optional)
 
-> If you are using a monorepo tool (Lerna, Nx), skip this section.
+> If you are using a monorepo tool (Lerna, Nx, etc.), skip this section.
```
alicewriteswrongs, line 179: this wasn't my experience, at least entering the commands as described in the tutorial right now.

I think that we'd have to modify the init commands to put both the Stencil component library and the Angular thingy in `packages/` in order for lerna to do this automatically. As is, I had to do `npm link` in the Stencil directory and then `npm link stencil-library` in the angular directory

src/docs/framework-integration/angular.md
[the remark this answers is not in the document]
alicewriteswrongs: Ok I wasn't able to get this to work until I added

```ts
export { DIRECTIVES } from './stencil-generated';
```

to `src/lib/component-library.component.ts`. Before then I got this error:

Screenshot 2023-01-05 at 2 25 30 PM

src/docs/framework-integration/angular.md
```diff
@@ -43,18 +47,28 @@ top-most-directory/
...
-npx -p @angular/cli ng new angular-workspace --no-create-application
-cd angular-workspace
-npx -p @angular/cli ng generate library component-library
+# From your top-most-directory/, initialize a workspace
+lerna init
+
+# install typescript and node types
+npm install typescript @types/node --save-dev
+# or if you are using yarn
+yarn add typescript @types/node --dev
```
alicewriteswrongs, line 71: I think when doing something like this it's probably a good idea to use a monorepo so you can do things like run an end-to-end build from one spot, manage publishing, etc.

However when following the guide at present we don't actually do much with the monorepo after initializing it, and it doesn't instruct the reader to put the two different projects (the stencil project and the angular project) into the `packages` subdirectory where lerna expects packages to be.

So I'm wondering, does it make sense to re-factor things so that the Stencil component library and the angular thingy are nested under `packages`, commands are run from the top-level project directory using `lerna`, etc? Or should we just scrap it, is there a standard, angular-ey way of doing this? not sure!

alicewriteswrongs: I noted an issue with linking below

alicewriteswrongs (changes requested):
a few things I noticed running through

src/docs/framework-integration/angular.md
alicewriteswrongs: I did `npm run watch` in the top-level of the angular workspace and that seems to have done what I want here, I think?

src/docs/framework-integration/angular.md
```diff
@@ -303,6 +367,92 @@ export const config: Config = {
...
+
+Event names shouldn’t include special characters when initially written in Stencil, try to lean on using camelCased event names for interoperability
+between frameworks.
+
+### How do I bind input events directly to a value accessor?
+
+You can configure how your input events can map directly to a value accessor, allowing two-way data-binding to be a built in feature of any of your
+components. Take a look at [valueAccessorConfig's option below]().
+
+### How do I add IE11 or Edge support?
```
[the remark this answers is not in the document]
tanner-reits: 👍 Yeah figured we'd tackle all the doc related deprecations at the same time

src/docs/framework-integration/angular.md
tanner-reits: Hmmm. We could. That just starts to get outside the bounds of Stencil and the wrapper and more into consuming the proxies in an Angular app. Not sure how far we wanna go down that path. Guess it depends who we're targeting with these docs: developers familiar with the framework being targeted, or developers who may have little to no experience with these frameworks. 

I'd argue that once we get outside the bounds of the wrappers, there are plenty of other docs available for getting a consuming app up and running. We could potentially link out to some resource like that if we want. Thoughts? @rwaskiewicz @alicewriteswrongs

src/docs/framework-integration/angular.md
tanner-reits: Yeah, I ran into this too actually with the React stuff. I'll make the same changes I had to here

src/docs/framework-integration/angular.md
rwaskiewicz: I do think there's a gap between this section and https://github.com/ionic-team/stencil-site/pull/951/files#diff-22dd5394aaba28bc0132da1cf756f97ec72e56ef3c64cab0c0bd0471550bc21eR209 (that existed prior to this commit).  While I agree we don't want to get into the business of describing every way to integrate with Angular in the docs, I do agree with @alicewriteswrongs that we should have a way to verify that each framework wrapper set up works. This is a decent sized pain point for new users, and I think is something we can improve on. I don't want us to get into the business of providing support for every Angular/React/Vue use case, but do think we should provide enough technical detail to say "Here is what you're doing and why. Here is an example that builds on that. Exercises in getting this working in other architectures are left to the reader (paraphrased)

alicewriteswrongs: ^^ my thoughts exactly, basically just after following the steps I was left wondering "did that all work?" and it is always good to leave a tutorial knowing you have something working to build on

tanner-reits: That "consumer usage" section still exists as it did before this re-work. But, I get where you guys are coming from. So, my plan is to add on to the "consumer usage" and add an MVP for creating an Angular app and ingesting the wrappers. I'll also see if anything like this exists for the other wrappers and add to those if needed as well

src/docs/framework-integration/angular.md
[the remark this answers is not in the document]
tanner-reits: That file is actually auto-generated. This did seem a little confusing, so I added some extra detail around which files are auto-generated, which can be deleted, and updated some naming to match what would be generated by following this doc step-by-step

tanner-reits:
@alicewriteswrongs @rwaskiewicz Alright, I made quite a few changes here based on y'alls feedback and some other things I ran into. Some of the stuff will need to be carried into the other targets as well, but gonna hold off on making those changes until this gets re-reviewed. I also successfully ran through this one step-by-step

src/docs/framework-integration/angular.md
```diff
@@ -70,15 +87,43 @@ npm install
...
+
+You will also need to add your generated Stencil library as a dependency so import references can be resolved correctly:
+
+```json
+// package.json
+
+"dependencies": {
+  "stencil-library": "*"
+}
+```
```
alicewriteswrongs, line 114: do I need to make this change in `packages/angular-workspace` or in `packages/angular-workspace/projects/component-library` ?

src/docs/framework-integration/angular.md
```diff
@@ -89,18 +134,31 @@ import { angularOutputTarget } from '@stencil/angular-output-target';
...
+> The `directivesProxyFile` is the relative path to the file that will be generated with all of the Angular component wrappers. You will replace the
+> file path to match your project's structure and respective names. You can generate any file name instead of `components.ts`.
 
-> The `directivesArrayFile` is the relative path to the file that will be generated with a constant of all the Angular component wrappers. This constant can be used to easily declare and export all the wrappers.
+> The `directivesArrayFile` is the relative path to the file that will be generated with a constant of all the Angular component wrappers. This
+> constant can be used to easily declare and export all the wrappers.
+
+> The `componentCorePackage` should match the `name` field in your Stencil project's `package.json`
+
+See the [API section below](#api) for details on each of the output target's options.
```
alicewriteswrongs, line 161: I have followed the steps recommended I believe exactly, but when I get to the `npm run build` below I'm getting a bunch of typescript errors preventing me from proceeding :/

Screenshot 2023-01-11 at 10 49 58 AM

alicewriteswrongs: here's the steps as I've been running them: 

https://gist.github.com/alicewriteswrongs/63533667b566925f2133dd50c66714d8

alicewriteswrongs (changes requested):
one thing I wasn't sure about, and then I'm facing a build issue w/ following the steps as recommended

src/docs/framework-integration/angular.md
tanner-reits: I actually ran into this once too. I think there was something weird going on with Jest peer deps. I tweaked the steps a bit so that shouldn't be an issue (at least for the purpose of this guide)

alicewriteswrongs: ok so I just updated the gist above (https://gist.github.com/alicewriteswrongs/63533667b566925f2133dd50c66714d8) with the steps in the guide. I'm still getting the same typescript error when I get to `npm run build` though :/

Is there a lerna related thing I'm not getting right here possible?

tanner-reits: Possibly. I wonder if it's related to `lerna add` I haven't been using those lerna commands since I think they're recommending moving away from that and using [package dependency management](https://lerna.js.org/docs/getting-started#package-dependency-management) which just takes advantage of npm/yarn workspaces. Not sure what `lerna add` really does under-the-hood if that is somehow causing dependency issues 🤷‍♂️

alicewriteswrongs: I don't think it's related to `lerna add` because I did the steps manually and replaced `lerna add` (where I have it in that script) with an edit to `packages/angular-workspace/packages/component-library/package.json` and I saw the same thing. Maybe it's something else about my machine...I'm not sure. I have been deleting all the output and then running through the steps with a clean slate each time though.

tanner-reits: Oh wait. This is an issue when building the Stencil project? I thought this was happening when building the Angular lib. I'll run through everything again and see if I hit this, but don't recall having an issue at that step

alicewriteswrongs: yeah exactly, when I get to the `npm run build` I get a type issue, it looks to me possibly like conflicting typedefs for Jest are being pulled in but I'm not sure

tanner-reits: Yep. It's a typedef conflict between Jest (which Stencil uses) and Jasmine (from Angular). Removing either solves the build problem. I'll add a step for that in the guide. Think I'll recommend removing Jasmine cuz Angular can work with Jest, but won't go beyond that since it isn't really the scope of the guide (we're not trying to completely architect a project)

alicewriteswrongs (approved):
things seem to be working alright now!

alicewriteswrongs:
do you maybe need to rebase this? it looks like it's not rebased over the Docusaurus changes yet

Screenshot 2023-01-26 at 10 25 22 AM

tanner-reits:
@alicewriteswrongs Yeah I need to rebase. @rwaskiewicz and I thought to first get all the doc reworks approved. Then I can merge them into one PR and handle the rebase to get all the different doc versions updated accordingly

alicewriteswrongs (approved):
lgtm!

src/docs/framework-integration/angular.md
```diff
@@ -89,193 +141,292 @@ import { angularOutputTarget } from '@stencil/angular-output-target';
...
+
 This section covers how developers consuming your Angular component wrappers will use your package and component wrappers.
 
-If you distributed your components through a primary `NgModule`, developers can simply import that module into their implementation to use your components.
+In order to use the generated component wrappers in the Angular app, you'll first need to build your Angular component library. From the root
+of your Angular workspace (`/packages/angular-workspace`), run the following command:
+
+```bash
+npx -p @angular/cli ng build component-library
+```
```
rwaskiewicz, line 287: When I get to this stage, I get an error:

Screenshot 2023-01-27 at 8 35 51 AM

Do I need to set this flag? Or maybe there's something I did wrong when adding `stencil-library` to my `package.json`? I just manually added it, w/o an `npm i` command, which I wouldn't think would be an issue 🤔

src/docs/framework-integration/angular.md
```diff
@@ -10,75 +10,127 @@ contributors:
...
+"dependencies": {
+  "stencil-library": "*"
+}
 ```
 
-### Adding Angular Output Target
+For more information, see the Lerna documentation on [package dependency management](https://lerna.js.org/docs/getting-started#package-dependency-management).
+
+:::note
+The Angular CLI will install Jasmine as a dependency to your Angular workspace. However, Stencil uses Jest as it's unit testing solution. To avoid
```
rwaskiewicz, line 122: Can we turn this into an `npm uninstall` command please? That way, we can use `npm2yarn` when this gets rebased (giving us the yarn equivalent more or less for free)

rwaskiewicz (changes requested):

src/docs/framework-integration/angular.md
tanner-reits: The `stencil-library` needed to be a peer dep instead of just a regular dep. I updated that step accordingly

src/docs/framework-integration/angular.md
```diff
@@ -89,193 +149,292 @@ import { angularOutputTarget } from '@stencil/angular-output-target';
...
+// app.module.ts
+
+import { ComponentLibraryModule } from 'component-library';
 
 @NgModule({
-  imports: [ExampleLibraryModule],
+  imports: [ComponentLibraryModule],
 })
-export class FeatureModule {}
+export class AppModule {}
```
rwaskiewicz, line 308: When I take this approach, and I put `my-component` in `app.component.html`, it doesn't appear to get registered with the custom elm registry. Does this `app.module.ts` look OK to you?
```
import { NgModule } from '@angular/core';
import { BrowserModule } from '@angular/platform-browser';

import { AppComponent } from './app.component';
import { ComponentLibraryModule } from 'component-library';

@NgModule({
  declarations: [
    AppComponent
  ],
  imports: [
    BrowserModule,
    ComponentLibraryModule
  ],
  providers: [],
  bootstrap: [AppComponent]
})
export class AppModule { }
```

src/docs/framework-integration/angular.md
```diff
@@ -89,193 +149,292 @@ import { angularOutputTarget } from '@stencil/angular-output-target';
...
+
+import { MyComponent } from 'component-library';
 
 @NgModule({
   declarations: [MyComponent],
-  exports: [MyComponent],
 })
-export class FeatureModule {}
+export class AppModule {}
 ```
```
rwaskiewicz, line 322: Is this example complete? I run into build errors with:
```tsx
import { NgModule } from '@angular/core';

import { AppComponent } from './app.component';
import { MyComponent } from 'component-library';
@NgModule({
  declarations: [
    AppComponent,
    MyComponent,
  ],
  providers: [],
  bootstrap: [AppComponent]
})
export class AppModule { }
```

> Error: projects/my-app/src/app/app.component.html:427:3 - error NG8001: 'my-component' is not a known element:
> 1. If 'my-component' is an Angular component, then verify that it is part of this module.
> 2. If 'my-component' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message.

rwaskiewicz (commented):
@tanner-reits I got closer! Just a few more snags

src/docs/framework-integration/angular.md
tanner-reits: @rwaskiewicz Thinking about this a little more, I think we should remove this section from the docs. This approach has some additional configuration/setup needed to make sure elements get registered correctly because they wouldn't have a mechanism in place to leverage the loader for registering/loading the custom elements. 

Pulling in the whole component library module is the approach we spend more time hashing out. Alternatively, this approach would work better when using the `dist-custom-elements` output. So, we could move some of this to the section in the docs here where we talk about that.

src/docs/framework-integration/angular.md
tanner-reits: @rwaskiewicz It looks like good to me. Did you add the initializer behavior which handles registering the custom elements?: https://github.com/ionic-team/stencil-site/blob/518a8619d865cef0bcd65b27697a78158f05bbb7/src/docs/framework-integration/angular.md?plain=1#L215-L235

rwaskiewicz (approved):

PR #958, opened 2022-12-19
docs(v3): format codeblock for install command

rwaskiewicz:
this commit changes the codeblock for installing `@stencil/core@v3-next` to use inline, monospaced font. this is due to a bug in the hydration of code blocks that include forward slashes. today, a code block with the original content, '@stencil/core@v3-next', may be rendered as:
```
@stencil/core@v3-nextcore@v3-next
```
where the content following the slash ('core@v3-next') is added as an additional text node in the DOM (without any line breaks).

with the upcoming migration to docusaurus for the stencil site, we choose to 'patch' this by moving away from the text block. it's slightly less readable, but ensures the install command is always correct
Screenshot 2022-12-19 at 10 16 05 AM


tanner-reits (approved):

alicewriteswrongs (approved):
sort of a sad bug! but let's do this for now

Issue #959, opened 2022-12-19
Docs should show @stencil/react-output-target is a dev dependency


postrad:
https://stenciljs.com/docs/react#install-the-react-output-target-in-your-stencil-component-library

Since the `@stencil/react-output-target` is a build-time compiler tool it should be a dev dependency.

PR #960, opened 2022-12-19
fix: React bindings should be installed as a dev dependency in example

postrad:
Since the `@stencil/react-output-target` is a build-time compiler tool it should be a dev dependency.

closes: https://github.com/ionic-team/stencil-site/issues/959

### DIFF
--- a/src/docs/framework-integration/react.md
+++ b/src/docs/framework-integration/react.md
@@ -75,13 +75,13 @@ If you do rename your React component library, be sure to change the `name` in t
 Now that the project structure is set up, we can install the React Output Target package in your Stencil component library. This package contains the React wrapper function that we will use to generate our React wrapped components inside a 'React component library'. To install the React Output Target package, run the following command in your **Stencil project directory**
 
 ```bash
-npm install @stencil/react-output-target
+npm install -D @stencil/react-output-target
 ```
 
 Or
 
 ```bash
-yarn add @stencil/react-output-target
+yarn add -D @stencil/react-output-target
 ```
 
 ### Add the React Wrapper Function to your Stencil Component Library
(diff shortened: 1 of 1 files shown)


src/docs/framework-integration/react.md
```diff
@@ -75,13 +75,13 @@ If you do rename your React component library, be sure to change the `name` in t
 Now that the project structure is set up, we can install the React Output Target package in your Stencil component library. This package contains the React wrapper function that we will use to generate our React wrapped components inside a 'React component library'. To install the React Output Target package, run the following command in your **Stencil project directory**
 
 ```bash
-npm install @stencil/react-output-target
+npm install -D @stencil/react-output-target
```
rwaskiewicz, line 78: Could we use the long-hand version of `-D` please? That way, should someone who isn't familiar with the CLI has a better chance of understanding what we're asking them to do.
```suggestion
npm install --save-dev @stencil/react-output-target
```

src/docs/framework-integration/react.md
```diff
@@ -75,13 +75,13 @@ If you do rename your React component library, be sure to change the `name` in t
...
 ```
 
 Or
 
 ```bash
-yarn add @stencil/react-output-target
+yarn add -D @stencil/react-output-target
```
rwaskiewicz, line 84: Could we use the long-hand version of `-D` please? That way, should someone who isn't familiar with the CLI has a better chance of understanding what we're asking them to do.
```suggestion
yarn add --dev @stencil/react-output-target
```

rwaskiewicz (changes requested):
Thanks! Had two minor suggestions

rwaskiewicz (approved):
Thanks!

rwaskiewicz:
No worries, thanks again!

PR #964, opened 2023-01-04
docs(react-integration): revamp into complete guide

tanner-reits:
Updates the docs for the React framework wrapper to follow the same structure as the Angular docs


src/docs/framework-integration/react.md
```diff
@@ -202,8 +218,153 @@ import { MyComponent } from 'component-library-react';
...
+
+### componentCorePackage
+
+**Optional**
+
+**Default: The `components.d.ts` file in the Stencil project's `package.json` types field**
+
+**Type: `string`**
+
+The title of the Stencil package where components are available for consumers (i.e. the `name` property value in your Stencil project's `package.json`).
```
rwaskiewicz, line 231: Optional word smithing to use some of the same nomenclature as the rest of the doc:
```suggestion
The name of the Stencil package where components are available for consumers (i.e. the value of the `name` property in your Stencil component library's `package.json`).
```

rwaskiewicz (changes requested):

src/docs/framework-integration/react.md
```diff
@@ -202,8 +218,153 @@ import { MyComponent } from 'component-library-react';
...
+
+### includeImportCustomElements
+
+**Optional**
+
+**Default: `undefined`**
+
+**Type: `boolean`**
+
+If `true`, the output target will import the custom element instance and register it with the Custom Elements Registry when the component is imported inside of a user's app. This can only be used with the [Custom Elements Bundle](/docs/custom-elements) and will not work with lazy loaded components.
```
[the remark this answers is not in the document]
tanner-reits: I think this is just an issue with the `dist-custom-elements` and `dist-custom-elements-bundle` target terminologies being used too loosely in the past. I'll update this to not say "Bundle" but keep the link the same. However, there may be some additional cleanup we need to address for V3 with the `loaderDir` config option since that is expecting a `defineCustomElements` function that would only exist if `customElementsExportBehavior: 'bundle'` is set.

src/docs/framework-integration/react.md
tanner-reits: @rwaskiewicz I like this change, but this same thing probably needs to happen in the Angular and Vue docs as well. So, I'll hold off on making this swap until we merge all these back into one PR so then I can just change them all at once

src/docs/framework-integration/react.md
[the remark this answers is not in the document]
rwaskiewicz: > I'll update this to not say "Bundle" but keep the link the same. 

👍 

> However, there may be some additional cleanup we need to address for V3 with the loaderDir config option since that is expecting a defineCustomElements function that would only exist if customElementsExportBehavior: 'bundle' is set.

Can you capture those TODOs somewhere so they don't fall through the cracks?

rwaskiewicz (changes requested):

versioned_docs/version-v2/framework-integration/react.md
```diff
@@ -12,199 +12,483 @@ contributors:
...
+:::note
+If you already have a React app, skip this section.
+:::
 
-```bash npm2yarn
-npm run build
-npm link
+From the `packages/` directory, run the following command to create a starter React app:
+
+```bash
```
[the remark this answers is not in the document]
tanner-reits: First off, I decided to swap this off of `create-react-app` to a Vite React app instead in [4d88bc3](https://github.com/ionic-team/stencil-site/pull/964/commits/4d88bc3a9b31d628edce650bdb62cbb66793c77d). Did this for two reasons:
1. The command for generating w/ `create-react-app` has some issues in Yarn environments
2. `create-react-app` is kinda dead, so it's easy enough for us to move away from it

Even so, we can't use `npm2yarn` here because that plugin can't correctly convert certain syntaxes (like `npm create`/`npm init`) to yarn syntax. It tries its best but then spits out a comment in the generated yarn block along the lines of "couldn't auto-convert command". I believe it's related to [this issue](https://github.com/facebook/docusaurus/issues/5861) in the docusaurus repo. I marked a TODO comment in the docs tho for us to look into using npm2yarn if/when that is ever resolved

rwaskiewicz (approved):
LGTM! 🚢

versioned_docs/version-v3.0/framework-integration/react.md
```diff
@@ -12,229 +12,506 @@ contributors:
...
+
+In addition, all the Web Component will be automatically defined as the generated component modules are bootstrapped.
+
+### What is the best format to write event names?
+
+Event names shouldn’t include special characters when initially written in Stencil. Try to lean on using camelCased event names for interoperability between frameworks.
+
+### How do I add IE11 or Edge support?
+
+If you want your custom elements to be able to work on older browsers, you should add the `applyPolyfills()` that surround the `defineCustomElements()` function.
```
alicewriteswrongs, line 509:
```suggestion
If you want your custom elements to work on older browsers, you should call the `applyPolyfills()` function before `defineCustomCustomElements()`. `applyPolyfills()` returns a `Promise`, so you could call it like this:
```

versioned_docs/version-v3.0/framework-integration/react.md
```diff
@@ -12,229 +12,506 @@ contributors:
...
+    })
+  ]
+}
+```
+
+Now, all generated imports will point to the default directory for the custom elements output. If you specified a different directory
+using the `dir` property for `dist-custom-elements`, you need to also specify that directory for the React output target. See
+[the API section](#customelementsdir) for more information.
+
+In addition, all the Web Component will be automatically defined as the generated component modules are bootstrapped.
```
alicewriteswrongs, line 501:
```suggestion
In addition, all the Web Components will be automatically defined as the generated component modules are bootstrapped.
```

versioned_docs/version-v3.0/framework-integration/react.md
```diff
@@ -12,229 +12,506 @@ contributors:
...
+### excludeComponents
+
+**Optional**
+
+**Default: `[]`**
+
+**Type: `string[]`**
+
+This lets you specify component tag names for which you don't want to generate React wrapper components. This is useful if you need to write framework-specific versions of components. For instance, in Ionic Framework, this is used for routing components - like tabs - so that
+Ionic Framework can integrate better with React's Router.
```
alicewriteswrongs, line 418: I think this should be 'React Router', React doesn't have a single routing solution (or a 'blessed' one):

```suggestion
Ionic Framework can integrate better with React Router.
```

docs/framework-integration/react.md
```diff
@@ -12,229 +12,506 @@ contributors:
...
+npm link name-of-your-stencil-package
 ```
 
-And with that, your component libraries are linked together. Now, you can make changes in your Stencil component library and run `npm run build` to propagate them through to the React component library without having to relink.
+The name of your Stencil package should match the `name` property from the Stencil component library's `package.json`.
+
+Your component libraries are now linked together. You can make changes in the Stencil component library and run `npm run build` to propagate the
+changes to the React component library.
 
 :::note
```
alicewriteswrongs, line 280: maybe should be

```suggestion
:::tip
```

docs/framework-integration/react.md
```diff
@@ -12,229 +12,506 @@ contributors:
...
-### Link Your Packages (Optional)
-
-If you want to build and test your components locally, you will need to link the packages together. This is a replacement for publishing packages to npm that allows you to develop and test locally. To do this, we’ll use the `npm link` command. This command creates a global symlink for a given package and thereby allows it to be consumed by other packages in your environment.
+### Registering Custom Elements
 
-First, build your Stencil component library. In your **Stencil component library**, run
+To register your web components for the lazy-loaded (hydrated) bundle, you'll need to expose a method for registering the underlying Stencil
+generated custom elements for the React proxy components to leverage. The easiest way to do this is to modify the React library's entry file
+to re-export the Stencil loader's `defineCustomElements()` method. In your React library's entry file (`packages/react-library/lib/index.ts`),
+add the follow:
```
alicewriteswrongs, line 246:
```suggestion
add the following:
```

alicewriteswrongs (changes requested):
looks good to me overall, just noticed a few quick text things

alicewriteswrongs (approved):
lgtm!

Issue #891, opened 2022-06-15
How to pass an Object to webcomponent base on stencil?


Linya-IronMan:
```tsx
@Component({
  tag: 'target-train-editor',
  styleUrl: 'target-train-editor.css',
  shadow: true,
})
export class TargetTrainEditor {
  @Prop({ attribute: 'train-resource', reflect: true, mutable: true }) trainResource?: string;

 render() {
    console.log(this.trainResource);
  }
}
```

```html
<target-train-editor train-resource="{}"></target-train-editor>
```
when the type of trainResource is "object" or "Record<string, any>", the console result is undefined,
when the type of trainResource is , the console result is {},

I also try the way in the doc
![image](https://user-images.githubusercontent.com/77786952/173830608-072af417-f4e1-48b8-aefd-89e800c9688c.png)

The result is also not as I wished;

Linya-IronMan:
How to resolve the problem, is there something wrong with my understanding of the doc?

rwaskiewicz:
Hey @Linya-IronMan 👋 

I apologize that it took so long for someone to acknowledge this issue. It is not possible to pass an object to a web component via HTML. I have created #973 to clarify this in our documentation

PR #973, opened 2023-01-12
docs(props): state limitation of attributes + objects

rwaskiewicz:
this commit updates the properties page to state that setting component's `@Prop()` decorated member that is of type `Object` cannot be set via an attribute

Closes: https://github.com/ionic-team/stencil-site/issues/891

### DIFF
--- a/src/docs/components/properties.md
+++ b/src/docs/components/properties.md
@@ -364,8 +364,44 @@ import { Component, Prop, h } from '@stencil/core';
     tag: 'todo-list-item',
 })
 export class ToDoListItem {
-    @Prop() labels: string[];
+    @Prop() itemLabels: string[];
 }
+```
+
+To set `itemLabels` in TSX, assign the prop name in the custom element's tag to the desired value like so:
+```tsx
+// TodoList.tsx
+import { Component, h } from '@stencil/core';
+import { MyHttpService } from '../MyHttpService';
+
+@Component({
+   tag: 'todo-list',
+   styleUrl: 'todo-list.css',
+   shadow: true,
+})
+export class ToDoList {
+   private labels = ['non-urgent', 'weekend-only'];
+
+   render() {
+      return <todo-list-item itemLabels={this.labels}></todo-list-item>;
+   }
+}
+```
+Note that the prop name is using `camelCase`, and the value is surrounded by curly braces.
+
+It is not possible to set `Array` props via an HTML attribute like so:
+```html
+<!-- this will not work -->
+<todo-list-item item-labels="['non-urgent', 'weekend-only']"></todo-list-item>
+```
+The reason for this is that Stencil will not attempt to serialize array-like strings written in HTML into a JavaScript object.
+Doing so can be expensive at runtime, and runs the risk of losing references to other nested JavaScript objects.
+
+Instead, properties may be set via `<script>` tags in a project's HTML:
+```html
+<script>
+   document.querySelector('todo-list-item').itemLabels = ['non-urgent', 'weekend-only'];
+</script>
 ```
 
 ### Advanced Prop Types
--- a/src/docs/components/properties.md
+++ b/src/docs/components/properties.md
@@ -300,6 +300,7 @@ import { MyHttpService } from '../path/to/MyHttpService';
     tag: 'todo-list-item',
 })
 export class ToDoListItem {
+    // Use `@Prop()` to declare the `httpService` class member
     @Prop() httpService: MyHttpService;
 }
 ```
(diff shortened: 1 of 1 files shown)


tanner-reits (approved):
looks good

src/docs/components/properties.md
```diff
@@ -312,9 +312,46 @@ export class MyHttpService {
...
+```
+Note that the prop name is using `camelCase`, and the value is surrounded by curly braces.
+
+It is not possible to set `Object` props via an HTML attribute like so:
+```html
+<!-- this will not work -->
+<todo-list-item http-service="{ /* implementation omitted */ }"></todo-list-item>
+```
+The reason for this is that Stencil will not attempt to serialize object-like strings written in HTML into a JavaScript object.
+Doing so can be expensive at runtime, and runs the risk of losing references to other nested JavaScript objects.
```
alicewriteswrongs, line 345: This in turn is part of a limitation of HTML attributes right? Might be good just to say something like

```suggestion
This is because HTML attribute values must be strings, and Stencil does not have support
for automatically deserializing JSON (or any other data interchange format) from attributes.
```

is that accurate?

alicewriteswrongs (changes requested):
some questions!

src/docs/components/properties.md
```diff
@@ -312,9 +312,46 @@ export class MyHttpService {
...
+   private httpService = new MyHttpService();
+
+   render() {
+      return <todo-list-item httpService={this.httpService}></todo-list-item>;
+   }
+}
+```
+Note that the prop name is using `camelCase`, and the value is surrounded by curly braces.
+
+It is not possible to set `Object` props via an HTML attribute like so:
```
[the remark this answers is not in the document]
rwaskiewicz: > should this possibly be in our little admonition? (the thing we mis-use blockquote for haha)

I thought about using the admonition/callout and the flow didn't seem quite right to me:
Screenshot 2023-01-18 at 8 17 23 AM
Screenshot 2023-01-18 at 8 16 27 AM



> Maybe also wording could be something like Note: non-primitive values cannot be passed to a custom attributes from HTML, only from JSX. So the following will not work:

I wanted to keep the language simple here. I see quite a bit on the Slack channel something to the effect of "Why can't I set a my object `@Prop` via HTML? When I do it through TSX it works!", and wanted the verbiage to match what folks might search (either via search engine or on the page itself with Ctrl+F)

src/docs/components/properties.md
```diff
@@ -312,9 +312,46 @@ export class MyHttpService {
 
 In TypeScript, `MyHttpService` is both an `Object` and a 'type'. When using user-defined types like `MyHttpService`, the
 type must always be exported using the `export` keyword where it is declared. The reason for this is Stencil needs to
-know what type the prop `myHttpService` is when passing an instance of `MyHttpService` to `TodoList` from a parent
+know what type the prop `httpService` is when passing an instance of `MyHttpService` to `TodoListItem` from a parent
 component.
 
+To set `httpService` in TSX, assign the prop name in the custom element's tag to the desired value like so:
```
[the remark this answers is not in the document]
rwaskiewicz: I'm not sure on this one - I don't think we necessarily need to call out non-primitives here, since we're in the `Object` section of the docs (and cover those types prior to this section).

If I remove that section calling out types of the class member decorated with `@Prop()`, we get:
```
Once we've used the @Prop() decorator to define a custom attribute, we can pass a value into our component in JSX within another Stencil component:
```

Which I think is roughly equivalent to what we have now? I do want to expand ['prop name' a bit to prevent muddying the differences between a property and an attribute](7658581d8) further though. LMK what you think!

src/docs/components/properties.md
rwaskiewicz: It's accurate, although Stencil tends to muddy the waters by serializing numbers represented as strings to `type number` in the JS (and TS) type system, and the same for booleans. 

Instead of getting into that (more than we have in the number/boolean sections), I wanted to call out a problem I've seen in Slack. Specifically, I've seen folks try to do something like:
```html
<to-do-list-item todos="{'thing': 'clean garage', 'isComplete': 'false'}"></to-do-list-item>
```
in HTML. Thoughts?

src/docs/components/properties.md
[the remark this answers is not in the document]
alicewriteswrongs: that makes sense, and yikes that really does not work in the admonition!

src/docs/components/properties.md
alicewriteswrongs: Ok gotcha, so maybe we could leave off any mention of HTML attributes needing to be strings, but call out something along the lines of 'Stencil does not have any support for deserializing objects from JSON' ?

rwaskiewicz: 👍 Done in 57415a573

alicewriteswrongs (approved):

PR #977, opened 2023-01-23
docs(v3-migration-guide): add section for framework wrappers

tanner-reits:
This commit adds a section to the v3 migration guide for the Stencil framework wrapper output targets. In this section, it talks about the impacts `customElementsExportBehavior` has on the framework wrappers and how to achieve the same behavior as in Stencil 2

### DIFF
--- a/src/docs/introduction/upgrading-to-stencil-three.md
+++ b/src/docs/introduction/upgrading-to-stencil-three.md
@@ -276,9 +276,7 @@ export const config: Config = {
 | `bundle`                      | A utility `defineCustomElements()` function is exported from the `index.js` file of the output directory. This function can be used to quickly define all Stencil components in a project on the custom elements registry.                                                                                                                                                                                 |
 | `single-export-module`        | All component and custom element definition helper functions will be exported from the `index.js` file in the output directory (see [Defining Exported Custom Elements](#defining-exported-custom-elements) for more information on this file's purpose). This file can be used as the root module when distributing your component library, see [below](#distributing-custom-elements) for more details. |
 
-:::note
-This option also has an impact when using the [Stencil framework integration output targets](/docs/overview). Please see the [framework integration migration](#framework-integration-output-targets) section below for more information.
-:::
+> **NOTE:** This option also has an impact when using the [Stencil framework integration output targets](/docs/overview). Please see the [framework integration migration](#framework-integration-output-targets) section below for more information.
 
 #### Move `autoDefineCustomElements` Configuration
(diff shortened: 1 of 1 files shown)


tanner-reits:
@rwaskiewicz @alicewriteswrongs This is for that small CS issue when they migrated to V3. Unfortunately, the framework targets basically expect the "single-export-module" behavior when `includeImportCustomElements: true` because it generates the same path when creating import statements for the `defineCustomElement` functions and all our general type declarations.

So, if we wanna provide alternative paths, I think we'd need to modify how those paths are generated in the output targets. Which might also include some more `package.json`/`stencil.config.ts` validation.

src/docs/introduction/upgrading-to-stencil-three.md
```diff
@@ -332,6 +334,36 @@ export const config: Config = {
...
 
+## Framework Integration Output Targets
+
+### Using `includeImportCustomElements`
+
+The addition of [customElementsExportBehavior](#add-customelementsexportbehavior-to-control-export-behavior) option on `dist-custom-elements` can also have an impact on the
+[Stencil framework integrations output targets](/docs/overview) if they are configured to use the output from `dist-custom-elements` (i.e. setting
+`includeImportCustomElements: true`). If problems are noticed with the generated import statements in the component proxy files, setting
+`customElementsExportBehavior: 'single-export-module'` in the framework output target config should result in the same behavior as in Stencil V2. For example, a project using
+the Angular framework integration output target would have a config like:
```
alicewriteswrongs, line 345: just some copy-edits

```suggestion
The addition of the [customElementsExportBehavior](#add-customelementsexportbehavior-to-control-export-behavior) option on `dist-custom-elements` can also have an impact on the
[Stencil framework integration output targets](/docs/overview) if they are configured to use the output from `dist-custom-elements` (i.e. by setting
`includeImportCustomElements: true`). If you encounter problems with the generated import statements in the component proxy files, setting
`customElementsExportBehavior: 'single-export-module'` in the `dist-custom-elements` output target config should result in the same behavior as in Stencil V2. For example, a project using
the Angular framework integration output target would have a config like:
```

alicewriteswrongs (changes requested):

rwaskiewicz:
@tanner-reits can you also open a separate PR with the same changes against `master` as well? I'd like to prevent a `git merge` at this point from `main->v3.0.0-dev` if possible

rwaskiewicz (approved):

alicewriteswrongs (approved):

PR #986, opened 2023-01-25
chore(scripts): delete scripts dir

alicewriteswrongs:
As far as I can tell this isn't being used for anything with the new Docusaurus-based build, so we might as well go ahead and delete it.


rwaskiewicz (approved):
LGTM @jaredcbaum can you please take a look and verify we don't need these anymore?

jaredcbaum (approved):

PR #992, opened 2023-01-25
fix(version-banner): banner matches container width/position

sean-perkins:
Updates the sizing and positioning of the version banner as seen on the "Next" version to match the main container sizing/positioning.

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -25,7 +25,11 @@
   }
 
   [class*='colorModeToggle'] {
-    margin-inline-start: -.25rem;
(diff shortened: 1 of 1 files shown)


src/css/custom.scss
```diff
@@ -25,7 +25,11 @@
   }
 
   [class*='colorModeToggle'] {
-    margin-inline-start: -.25rem;
+    margin-inline-start: -0.25rem;
```
sean-perkins, line 28: This is from prettier plugin in VSCode.. I can revert if we care.

tanner-reits (approved):
Thank you!

PR #999, opened 2023-01-25
fix(docs-hiding): update logic for excluding docs

rwaskiewicz:
this commit updates the logic for excluding docs to account for docs that are in versioned docs directories. prior to this commit, docs in the v2 and v3.0 directories wouldn't be excluded, causing for an incorrect sidebar to be generated

## Before
Screenshot 2023-01-25 at 4 58 01 PM

## After
Screenshot 2023-01-25 at 4 57 49 PM


alicewriteswrongs (approved):
seems to work!

tanner-reits (approved):

PR #1004, opened 2023-01-26
docs(component): port component lifecycle chart to React

alicewriteswrongs:
We previously had an interactive chart of the different component lifecycle methods on the corresponding page which was implemented as a Stencil component. This ports that component to React, making the minimal set of changes which I believe are necessary to get it working.

### TESTING

Check that:

- you can see the chart on the `docs/component-lifecycle` page for all three versions
- the hovering styling effect works as it used to in the Stencil-based implementation
- the links in the chart all point to the right page (i.e. the link to the `@Watch` documentation should point to that page for the version that you're currently looking at)

### DIFF
--- a/docs/components/component-lifecycle.md
+++ b/docs/components/component-lifecycle.md
@@ -13,8 +13,9 @@ Components have numerous lifecycle methods which can be used to know when the co
 
 Implement one of the following methods within a component class and Stencil will automatically call them in the right order:
 
-<lifecycle-chart></lifecycle-chart>
+import LifecycleMethodsChart from '@site/src/components/LifecycleMethodsChart';
 
+<LifecycleMethodsChart />
 
 ## connectedCallback()
 
--- a/src/components/LifecycleMethodsChart.module.css
+++ b/src/components/LifecycleMethodsChart.module.css
@@ -1,0 +1,27 @@
+.lifecycleChart {
+  display: block;
+
+  --opacity: 1;
+}
+
+.lifecycleChart.highlight {
+  --opacity: 0.5;
+}
+
+.lifecycleChart rect,
+.lifecycleChart text,
+.lifecycleChart path {
+  opacity: var(--opacity);
+  transition: opacity 0.2s ease-in-out;
+}
+
+.lifecycleChart .trigger {
+  cursor: default;
+}
+
+.lifecycleChart.highlight-init .pathInit,
+.lifecycleChart.highlight-attached .pathAttached,
+.lifecycleChart.highlight-removed .pathRemoved,
+.lifecycleChart.highlight-update .pathUpdate {
+  --opacity: 1;
+}
--- a/src/components/LifecycleMethodsChart.tsx
+++ b/src/components/LifecycleMethodsChart.tsx
@@ -1,0 +1,5 @@
+import React from 'react'
+
+export default function LifecycleMethodsChart () {
+  return <h1>LIFECYCLE</h1>
+}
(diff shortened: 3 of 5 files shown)


docs/components/component-lifecycle.md
```diff
@@ -13,8 +13,9 @@ Components have numerous lifecycle methods which can be used to know when the co
 
 Implement one of the following methods within a component class and Stencil will automatically call them in the right order:
 
-<lifecycle-chart></lifecycle-chart>
+import LifecycleMethodsChart from '@site/src/components/LifecycleMethodsChart';
 
+<LifecycleMethodsChart />
```
alicewriteswrongs, line 18: because these files are processed by Docusaurus as 'mdx' files we can just import a React component and render them as if we were in a `.jsx` or `.tsx` file (we don't need to `import React from 'react'` however)

I believe that there are some differences between the `md` and `mdx` parsers, and if you have anything it believes is JSX it will switch to using the `mdx` one. You can also force this by using an `.mdx` file extension.

More info here: https://docusaurus.io/docs/markdown-features/react

src/components/LifecycleMethodsChart.module.css
```diff
@@ -0,0 +1,27 @@
...
+.lifecycleChart .trigger {
+  cursor: default;
+}
+
+.lifecycleChart.highlight-init .pathInit,
+.lifecycleChart.highlight-attached .pathAttached,
+.lifecycleChart.highlight-removed .pathRemoved,
+.lifecycleChart.highlight-update .pathUpdate {
+  --opacity: 1;
+}
```
alicewriteswrongs, line 27: This is all copied from the Stencil-based implementation, with the only changes some `snake-case -> camelCase` stuff and the like.

src/components/LifecycleMethodsChart.tsx
```diff
@@ -0,0 +1,198 @@
+import React, { useState } from 'react';
+import Link from '@docusaurus/Link';
```
alicewriteswrongs, line 2: We need to use Docusaurus' `Link` component for any links we do within a React component. This lets us take advantage of some pre-fetching stuff that the component does automatically, and it also ensures that we can utilize Docusaurus' version-aware relative link resolution.

In particular, if we do something like this:

```jsx
<Link to="reactive-data#watch-decorator">@Watch()</Link>
```

That will render as a link to `/docs/v2/reactive-data#watch-decorator`, `/docs/reactive-data#watch-decorator`, or `/docs/next/reactive-data#watch-decorator`, depending on which version of the page we're currently on.

alicewriteswrongs: See here for more info: https://docusaurus.io/docs/docusaurus-core#link

alicewriteswrongs (commented):
just noting a few things

docs/components/component-lifecycle.md
```diff
@@ -13,8 +13,9 @@ Components have numerous lifecycle methods which can be used to know when the co
 
 Implement one of the following methods within a component class and Stencil will automatically call them in the right order:
 
-<lifecycle-chart></lifecycle-chart>
+import LifecycleMethodsChart from '@site/src/components/LifecycleMethodsChart';
```
tanner-reits, line 16: Probably a dumb question, but how does this import get resolved? Are we defining this path somewhere, or is this some docusaurus magic?

alicewriteswrongs: I believe it is magic 🪄💫🌟

I followed what it recommends here: https://docusaurus.io/docs/markdown-features/react#importing-components

I think it injects this `@site` scope into the build somehow

rwaskiewicz (approved):

PR #1005, opened 2023-01-26
fix(browser-support): restore styling of browser-support page

alicewriteswrongs:
this is a quick-and-dirty restoration of the styling on the browser support page, where we had some custom stuff before. The web team should go over this in more detail at some point, but this at least gets us away from some unstyled text that is there atm.

### DIFF
--- a/docs/reference/browser-support.md
+++ b/docs/reference/browser-support.md
@@ -18,19 +18,19 @@ Stencil builds Web Components that run natively or near-natively in all widely u
   <div class="bs-chart__group">
     <div class="bs-chart__cards">
       <div class="bs-chart__card">
-        <app-icon name="checkmark"></app-icon>
+        <ion-icon name="checkmark"></ion-icon>
         Chrome 79+
       </div>
       <div class="bs-chart__card">
-        <app-icon name="checkmark"></app-icon>
+        <ion-icon name="checkmark"></ion-icon>
         Safari 14+  
       </div>
       <div class="bs-chart__card">
-        <app-icon name="checkmark"></app-icon>
+        <ion-icon name="checkmark"></ion-icon>
         Firefox 70+
       </div>
       <div class="bs-chart__card">
-        <app-icon name="checkmark"></app-icon>
+        <ion-icon name="checkmark"></ion-icon>
         Edge 79+
       </div>
     </div>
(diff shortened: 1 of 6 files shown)


src/css/custom.scss
```diff
@@ -32,4 +32,80 @@
...
+
+  .bs-chart,
+.bs-chart__cards,
+.bs-chart__card {
+  display: flex;
+}
+
+.bs-chart {
+  margin: 40px 0;
+  justify-content: space-between;
```
rwaskiewicz, line 44: I think with 'space-between', this ends up looking a little funky:

Screenshot 2023-01-26 at 1 03 27 PM

I think 'space-between' in addition to the content for IE11 being left justified in its container is giving it a bit of an off look.

If we switch this to 'center':
```suggestion
  justify-content: center;
```
Screenshot 2023-01-26 at 1 06 44 PM

That doesn't fix the text though 🤔

docs/reference/browser-support.md
```diff
@@ -18,19 +18,19 @@ Stencil builds Web Components that run natively or near-natively in all widely u
   <div class="bs-chart__group">
     <div class="bs-chart__cards">
       <div class="bs-chart__card">
-        <app-icon name="checkmark"></app-icon>
+        <ion-icon name="checkmark"></ion-icon>
```
rwaskiewicz, line 21: I didn't know we were pulling in Ionic Framework already 🤔

alicewriteswrongs:
ah sorry should have marked as draft! Still cleaning a few things up

docs/reference/browser-support.md
alicewriteswrongs: yeah we are currently pulling in ionicons here:

https://github.com/ionic-team/stencil-site/blob/40fd73cba2c109baa1c45ed8fb4d7bdc9bf7acc6/docusaurus.config.js#L28-L33

alicewriteswrongs: I thought, however, that since we use emojis for charts like this on e.g. the [support policy](https://stenciljs.com/docs/next/support-policy) page we might want to make that change on this page too - I did that so we can see what we think for the polyfills chart at the bottom of the page

src/css/custom.scss
alicewriteswrongs: I already removed `space-between` in favor of not setting it, but I think `center` looks a little nicer so I'll add that back

the text is a little funny, but I'm alright with that for now if you are haha

alicewriteswrongs: Screenshot 2023-01-26 at 1 30 03 PM

^^ this is what it looks like with `justify-content: center` -- IE is sort of set off to the side haha

docs/reference/browser-support.md
alicewriteswrongs: Screenshot 2023-01-26 at 1 21 14 PM

looks like this

rwaskiewicz (approved):

tanner-reits (approved):
lol much better than what's there now!

PR #1017, opened 2023-02-01
chore(github): add dependabot to stencil docs

rwaskiewicz:
this commit adds a dependabot configuration to the stencil site to check for updated packages used by the site itself.

at the time of this commit, the ionic-preset-classic project was intentionally excluded from this configuration, as we expect it to be a separate npm package in the near future.

this configuration checks weekly (as oppposed to daily in other stencil projects) in order to see how a weekly-based notification might help balance providing the team more time to work on stencil, rather than dependabot updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: 'npm' # See documentation for possible values
+    directory: '/' # Location of package manifests
+    schedule:
+      interval: 'weekly'
+    open-pull-requests-limit: 50
+    # Disable rebasing for pull requests, as having several open pull requests all get simultaneously rebased gets
+    # noisy from a notification standpoint
+    rebase-strategy: 'disabled'


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+version: 2
+updates:
+  - package-ecosystem: 'npm' # See documentation for possible values
+    directory: '/' # Location of package manifests
+    schedule:
+      interval: 'weekly'
+    open-pull-requests-limit: 50
+    # Disable rebasing for pull requests, as having several open pull requests all get simultaneously rebased gets
+    # noisy from a notification standpoint
+    rebase-strategy: 'disabled'
```
rwaskiewicz, line 13: @jaredcbaum this will lead to opening a new PR for any dependency that is not up to date in the project's root level `package.json`. Are there any (e.g. React) that we should _not_ update/do manually on a given cadence? If so, I can add an `ignore` list similar to https://github.com/ionic-team/stencil/blob/dd450aa6f2d2b208f01f195c7cdb1231b7ff635e/.github/dependabot.yml#L15 here. Just LMK what packages/versions we shouldn't update

jaredcbaum: Nothing comes to mind after looking through the deps.

jaredcbaum: Even if so, I'm sure we can fine tune after the fact.

jaredcbaum (approved):

.github/dependabot.yml
rwaskiewicz: 👍 Is there a max version of React I should put in here? e.g. No higher than 17.Y.Z? 17.0.Z?

alicewriteswrongs (approved):

.github/dependabot.yml
jaredcbaum: https://github.com/ionic-team/stencil-site/blob/main/ionic-preset-classic/package.json#L37

tanner-reits (approved):

PR #1035, opened 2023-02-08
Update 03-getting-started.md

SavvasJacksonGee:
Correct typo on line 89 from 'my-first-stencil project' to 'my-first-stencil-project' . Stencil project names must be kebab case and cannot include spaces.


rwaskiewicz (approved):
Thanks! This change also needed to be added to the Stencil v2 and 'next' versioned docs, which I went ahead and updated.

PR #1037, opened 2023-02-09
chore(docusaurus): remove unneeded readme files

rwaskiewicz:
this commit removes readme files that were once used in the stencil site when the site was based on stencil's ssr infra. previously, a single readme file would be used to generate the lefthand sidebar for navigation purposes through a series of build scripts. when we moved over to docusaurus, that file came along for the ride, and was copied to multiple directories when versioned docs were added to the site. this commit removes those files, as well as the exclude rule for each of them in the docusaurus config


alicewriteswrongs (approved):
lgtm! 🗑️

tanner-reits (approved):

PR #1040, opened 2023-02-16
style(theme): No striped tables

jaredcbaum:
This will make striped rows of tables the same background as unstriped rows.


rwaskiewicz (approved):
Thanks Jared, looks great!

PR #1046, opened 2023-02-20
feat(config): Add documentation of `transformAliasedImportPaths` option

alicewriteswrongs:
This adds documentation of the new `transformAliasedImportPaths` option, added to Stencil in https://github.com/ionic-team/stencil/pull/4042/.

### DIFF
--- a/docs/config/01-overview.md
+++ b/docs/config/01-overview.md
@@ -273,3 +273,57 @@ taskQueue: 'async'
 ## testing
 
 Please see the [testing config docs](../testing/config.md).
+
+## transformAliasedImportPaths
+
+*default: `false`*
+
+This sets whether or not Stencil should transform [path aliases](
+https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) set
+in a project's `tsconfig.json` from the assigned module aliases to resolved
+relative paths. This will not transform external imports (like `@stencil/core`) or
+relative imports (like `'../utils'`).
+
+This option applies globally and will affect all code processed by Stencil,
+including `.d.ts` files and spec tests.
+
+An example of path transformation could look something like the following.
+
+First, a set of `paths` aliases in `tsconfig.json`:
+
+```json title="tsconfig.json"
+{
+  "compilerOptions": {
+    "paths": {
+      "@utils": [
+        "../path/to/utils"
+      ]
+    }
+  }
+}
+```
+
+Then with the following input:
+
+```ts title="src/my-module.ts"
+import { utilFunc, UtilInterface } from '@utils'
+
+export function util(arg: UtilInterface) {
+    utilFunc(arg)
+}
+```
+
+if the `transformAliasedImportPaths` option is set to `true` Stencil will
+produce the following output:
+
+```js title="dist/my-module.js"
+import { utilFunc } from '../path/to/utils';
+export function util(arg) {
+    utilFunc(arg);
+}
+```
+
+```ts title="dist/my-module.d.ts"
+import { UtilInterface } from '../path/to/utils';
+export declare function util(arg: UtilInterface): void;
+```


docs/config/01-overview.md
```diff
@@ -273,3 +273,37 @@ taskQueue: 'async'
 ## testing
 
 Please see the [testing config docs](/testing-config).
+
+## transformAliasedImportPaths
```
rwaskiewicz, line 277: Once we land this flag/feature, do you think we should add test path alias transformations to this flag, or introduce another one?  I'm wondering if we should add a comment about these paths for spec tests in this section a) in general, b) for now, c) not at all. Thoughts?

docs/config/01-overview.md
```diff
@@ -273,3 +273,37 @@ taskQueue: 'async'
...
+This sets whether or not Stencil should transform [path aliases](
+https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) set
+in a project's `tsconfig.json` from the assigned module aliases to resolved
+relative paths. This will not transform external imports (like `@stencil/core`) or
+relative imports (like `'../utils'`).
+
+This option applies globally and will affect all code emitted by Stencil,
+including `.d.ts` files.
+
+An example of path transformation could look something like:
```
rwaskiewicz, line 290: Consider adding a small snippet of a `tsconfig.json` to demonstrate paths:
```json title="tsconfig.json"
{
    "paths": {
      "@utils": [
        "../path/to/utils"
      ]
    }
}
```

docs/config/01-overview.md
```diff
@@ -273,3 +273,37 @@ taskQueue: 'async'
...
+
+```ts
+// Source code (ts)
```
rwaskiewicz, line 293: FWIW - this could be combined into one line with the `title` attribute:
```suggestion
```ts title="Source Code (.ts)"
```
But, that would require us to break this out into three separate blocks:
1. Source Code
2. Output Code
3. Type Declaration

docs/config/01-overview.md
alicewriteswrongs: My instinct would be to keep it simple and not have a separate flag for test behavior, so that if you set this to `true` you'd get the transformer running on spec tests as well.

I was also hoping to land the change for spec tests in 3.1 alongside this change, but I can add a small 'info' thingy now saying that this doesn't yet apply to spec tests in case we end up shipping the change before the patch for spec tests lands.

docs/config/01-overview.md
alicewriteswrongs: ooo I forgot about this functionality w/ docusaurus!

alicewriteswrongs: I turned it into three codeblocks, added a little more inline explanation, and added a `tsconfig.json` example as well

docs/config/01-overview.md
rwaskiewicz: > I was also hoping to land the change for spec tests in 3.1 alongside this change, but I can add a small 'info' thingy now saying that this doesn't yet apply to spec tests in case we end up shipping the change before the patch for spec tests lands.

Yeah, that SGTM 👍

rwaskiewicz (approved):

rwaskiewicz:
@tanner-reits I think so. Since 'next' will turn into 3.1 in the docs (and that flag isn't available in 3.0)

tanner-reits:
@rwaskiewicz Ohhh right right right because we're doing versions for minor releases as well. Makes sense, thanks!

alicewriteswrongs:
just updated this to indicate that the transform works in spec tests after merging https://github.com/ionic-team/stencil/pull/4090

rwaskiewicz (approved):

tanner-reits (approved):

PR #1047, opened 2023-02-22
chore(repo): add prettier infrastructure

rwaskiewicz:
this commit adds prettier infrastructure to the project. prettier was
previously a dependency of the project, but was not being used. similar
to other stencil projects, this commit uses ionic's prettier
configuration for consistent formatting.

three npm scripts have been added for prettier:
1. `prettier.base` to serve as a single source of truth for common
   settings, such as directories to format
2. `prettier.dry-run` to act as a non-mutating prettier run. used in ci
3. `prettier` to format and write changes back to disk

to start, we shall only format configuration scripts (JavaScript +
TypeScript). at the time of this writing, there are:
- two previous major iterations of the docs (v2, v3)
- nine minor iterations for stencil v4
- the `next` section of the documentation (slated for stencil v4.10.0)
at this time, i would prefer to minimize churn related to formatting
these 12 directories (2 + 9 + 1). perhaps shortly after the release of
stencil v5 (where the v4 documentation is collapsed into one directory)
will we consider either formatting all major version directories, or
just the v5 directory.

prettier and the ionic prettier configuration dependencies have been
moved to `dev-dependencies`, as they are not needed in the production
build of the site.

a ci workflow (`main.yml`) has been added to kick off a dry-run of
prettier. should any files not be properly formatted, the workflow shall
fail. the intent is to block landing code on this check in github.

the second commit runs the formatter.


package.json
```diff
@@ -13,22 +13,25 @@
...
-    "spellcheck": "cspell --no-progress \"docs/**/*.md\" \"versioned_docs/**/*.md\" \"*.md\""
+    "spellcheck": "cspell --no-progress \"docs/**/*.md\" \"versioned_docs/**/*.md\" \"*.md\"",
+    "prettier": "npm run prettier.base -- --write",
+    "prettier.base": "prettier --cache \"./*.(js|jsx|ts|tsx|json5|json)\"",
+    "prettier.dry-run": "npm run prettier.base -- --list-different"
   },
   "dependencies": {
     "@docusaurus/core": "^3.0.1",
     "@docusaurus/remark-plugin-npm2yarn": "^3.0.1",
-    "@ionic/prettier-config": "^4.0.0",
```
rwaskiewicz, line 21: Just moving these to `devDependencies` here

christian-bromann (approved):
Nice 👌

jaredcbaum (approved):

tanner-reits (approved):

PR #1053, opened 2023-02-23
fix(landing): update slug for landing page

rwaskiewicz:


rwaskiewicz:
This isn't the right fix, closing

PR #1067, opened 2023-03-15
fix wrong import guidance

FirstVertex:
it is not possible to import something that has a name with a dash.  it's not accurate to suggest to import without curly braces.  it's incomplete to not show the "from" clause of the import statement.


christian-bromann:
Thanks for raising the PR, I re-raised this PR and applied the change to other doc versions, see https://github.com/ionic-team/stencil-site/pull/1469. I will go ahead and close this one.

PR #1068, opened 2023-03-16
chore(docs): remove unused frontmatter 

rwaskiewicz:
this commit removes unused 'contributor' frontmatter from the site.
prior to switching to docusaurus, it was used as a part of the stencil
site build to generate 'contributor bubbles' at the bottom of the page.
we no longer do this, and if we wanted to could derive this data from
git history (at the cost of calculating it).

remove it as it's unclear for new contributors if they should update
these or not.

remove it as it also conflicts with new cspell integration work, which
flags many github user names as mis-spellings

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ When it comes time to release a new minor version of Stencil, the contents of `.
    1. For example, if your doc is a guide for Stencil v3.0, you would put it in `./versioned_docs/version-v3.0/guides`.
    2. Depending on the nature of the documentation, the Stencil team may wish to "port" this change to other versions of the Stencil documentation. During a review of your PR, the team will be able to give guidance how to propagate this change.
 2. Write your documentation following the style in the other docs markdown files. Try to aim for being as clear and concise as possible. We recommend checking out the [vue.js docs](https://vuejs.org/) for examples of good docs.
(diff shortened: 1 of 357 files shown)


rwaskiewicz:
Unsure why the build is broken, as it works locally :shrug: marking as draft

CONTRIBUTING.md
```diff
@@ -60,7 +60,7 @@ When it comes time to release a new minor version of Stencil, the contents of `.
    1. For example, if your doc is a guide for Stencil v3.0, you would put it in `./versioned_docs/version-v3.0/guides`.
    2. Depending on the nature of the documentation, the Stencil team may wish to "port" this change to other versions of the Stencil documentation. During a review of your PR, the team will be able to give guidance how to propagate this change.
 2. Write your documentation following the style in the other docs markdown files. Try to aim for being as clear and concise as possible. We recommend checking out the [vue.js docs](https://vuejs.org/) for examples of good docs.
-3. Make sure the page header contains title, description, url and contributors. See other docs files for examples.
+3. Make sure the page header contains title, description, and url. See other docs files for examples.
```
tanner-reits, line 63: We should probably change this to `slug` instead of `url` while we're here

tanner-reits (changes requested):
One small tweak. Otherwise should be good once CI issues are resolved

alicewriteswrongs (approved):
lgtm!

PR #1071, opened 2023-03-16
Fix/versioned redirects

jaredcbaum:
This PR adds functionality to the vercel.config file to include versioned docs for the current redirects. In addition, it removes pattern matching for trailing slashes and adds a config option to remove them by default.

Here is the current [pattern matching library](https://github.com/pillarjs/path-to-regexp/tree/v6.1.0) that vercel uses for its config file. Here is the [related section](https://vercel.com/docs/errors#invalid-route-source-pattern) in the vercel docs.

Unfortunately this is a blanket solution that will affect all versions (including current) of certain files, some which may not want redirects or require different destinations based on versions. I'm open to suggestions on how to mitigate this. ATM the best solution I can think of is to just handle those on a case by case basis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,8 @@ www
 # IDEs
 .idea/
 .vscode/
-.vercel
 
 # vercel
+.vercel
 auth.json
 config.json
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,8 @@ www
 # IDEs
 .idea/
 .vscode/
+.vercel
+
+# vercel
+auth.json
+config.json
--- a/vercel.json
+++ b/vercel.json
@@ -2,142 +2,142 @@
   "trailingSlash": false,
   "redirects": [
     {
-      "source": "/:version(v[\\d]+(?:\\.[\\dx]+)?)?",
+      "source": "/:version(next|(?:v[\\d]+(?:\\.[\\dx]+))?)?",
       "destination": "/docs/:version/introduction",
       "statusCode": 301
     },
     {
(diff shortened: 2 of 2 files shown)


jaredcbaum:
I'm am going to add functionality to make this work for links with `/next/` in the url as well

jaredcbaum:
In order to see this working on the preview url, you will have to use `/` as the base url instead of `/docs` since this deployment is being rewritten to `stenciljs.com/docs`

rwaskiewicz (changes requested):
Thanks @jaredcbaum!

At the moment, I don't think the regex's handle the v2 version, which doesn't have a trailing period + numeric character(s) at the end. To test this, I went to the preview site and transformed the URL like so:
```diff
- https://stencil-docs-git-fix-versioned-redirects-ionic1.vercel.app/docs/v2/introduction
+ https://stencil-docs-git-fix-versioned-redirects-ionic1.vercel.app/v2/
```
where I 
- removed `docs/` per the GH summary instructions
- remove `introduction', which so that the resulting URL should redirect us to the intro page

However, when I do that, I get a 404 page. If I do the same with a v3.x URL:
```diff
- https://stencil-docs-git-fix-versioned-redirects-ionic1.vercel.app/docs/v3.0/introduction
+ https://stencil-docs-git-fix-versioned-redirects-ionic1.vercel.app/v3.0/
```
it properly redirects

rwaskiewicz (approved):

rwaskiewicz:
Thanks!

PR #1090, opened 2023-04-07
docs(docs): create a separate section for doc generation

rwaskiewicz:
⚠️ this should be considered a proposal, we can absolutely reject this ⚠️ 

this commit (pr) separates the documentation generation section from the output targets section of the docs. the primary motivation for this change is to surface the documentation generation aspects of stencil in our doc site by creating a separate section in the side bar for it.

as you can see, when we separate these, it reveals that both sections are a little lacking in terms of substance! today, we dump a lot of data in the "output targets" section and let that volume make up for quality content. we'll have to work on improving both sections of the docs. this is one potential first step.

no redirects are necessary, since we don't have routing that is strictly related to the our directory structure of the project anymore 😎  

assuming this is something we _want_ to do, I'll go ahead a propagate these changes to all 3.x branches

Before:
Screenshot 2023-04-07 at 10 11 47 AM
After:
Screenshot 2023-04-07 at 10 12 11 AM


rwaskiewicz:
@JessicaSachs I added you for general viz, no pressure to review while you're onboarding

tanner-reits:
I don't have any concerns or objections with this. It'd be nice if we could nest this under "output targets", but that would negate surfacing this for visibility. 

One side note semi-related to this is the distinction between `custom` and `docs-custom`. I don't believe we document the `custom` target anywhere. But, if we wanted to change that, we'd need to add clarity around their use-cases since `custom` accepts a `docs` param for its `generator` function.

docs/documentation-generation/01-overview.md
```diff
@@ -0,0 +1,43 @@
...
+
+- [`docs-readme`: Documentation readme files formatted in markdown](./docs-readme.md)
+- [`docs-json`: Documentation data formatted in JSON](./docs-json.md)
+- [`docs-custom`: Custom documentation generation](./docs-custom.md)
+- [`docs-vscode`: Documentation generation for VS Code](./docs-vscode.md)
+- [`stats`: Stats about the compiled files](./docs-stats.md)
+
+## Docs Auto-Generation
+
+As apps scale with more and more components, and team size and members continue to adjust over time, it's vital all components are well documented, and that the documentation itself is maintained. Maintaining documentation is right up there with some of the least interesting things developers must do, but that doesn't mean it can't be made easier.
```
alicewriteswrongs, line 18: I sort of like writing documentation / technical things! It's a great way to clarify ideas and present how something works - I'd kind of like it if we could remove / change the wording here that suggests it's boring or a chore to write documentation — I know many developers do feel that way though, so maybe we could say something more like

"Whether you love taking the time to write documentation or you can't wait to get coding again, we could all use some help in keeping documentation up-to-date when we change components and APIs. Stencil provides some built-in tools to automate extracting information from your component library, providing flexible output formats that you can use when building out your library's documentation."

Idk if that quite rolls of the tongue, but something along those lines?

alicewriteswrongs (changes requested):
One question about the text while we're in here (I know this is only showing up as green as an artifact of the move)

I like the move overall and think it will surface documentation generation as a feature of Stencil more to the reader!

Unrelated to whether this is a good idea or not, I want to point out a little inconsistency in our sidebar:

Screenshot 2023-04-12 at 10 09 42 AM

Basically one of the documentation output targets has a page title which looks like a title ("VS Code Documentation") while the others have the names of the output targets ("docs-json", "docs-custom", etc).

Should we standardize one way or the other?

Personally I think I might find it a little more inviting if they were titled like so:

Screenshot 2023-04-12 at 10 12 20 AM

I don't know what the exact titles are there, but it might be nice to have something like that, especially because after moving these out from under the "Output Targets" heading it may not be clear to the reader that a title like `docs-json` is the output target type that you'd use for doing JSON documentation output.

Another way to put this is that if they're living in a separate section for Documentation Generation, then conceptually those pages are no longer documentation _for specific output targets_ but instead generally covering how to use Stencil to generate, say, documentation in the JSON format. The fact that that is done w/ an output target, configured in the Stencil configuration, is something of an implementation detail — maybe in the future we'd have a page here for something which is not output target-driven but instead used via the CLI, for instance.

In summary, thank you for coming to my TED talk about how to overthink small changes 😅

rwaskiewicz:
Does anyone have any qualms with us _just_ updating `next` and `v3.2`? I wonder if a refactor of this size (including updating titles, copy, etc.) should just be limited to the latest versions of the docs

alicewriteswrongs:
yeah I think that makes sense, we can make changes like this and have them apply going forward, but no need to rewrite history :)

tanner-reits (approved):
I'm onboard to only change 3.2 and next. Deployment looks good in that regard 👍

alicewriteswrongs (approved):
thought I had already approved, sorry about that!

PR #1100, opened 2023-04-26
fix(links): update links to stencil slack

rwaskiewicz:
this commit updates the links to the stencil slack. the heroku app is no longer working/maintained. to ensure that links continue to work, switch to using the slack link directly.

I backported this to all versions, as:
1. I think it's possible that folks using earlier versions would expect the Slack link to not change/be available
2. It actually makes it easier when searching through the `ionic-team` org in GH to find repos using this old link

### Testing

The links in the sidebar and FAQ section have been tested in the Vercel preview across all versions to ensure they work.

Screenshot 2023-04-26 at 9 41 41 AM
Screenshot 2023-04-26 at 9 41 19 AM


alicewriteswrongs (approved):
looks good!

PR #1116, opened 2023-05-17
docs(output-targets): enabling & configuring target validation

tanner-reits:
Adds documentation for the multi-target field validation. This does _not_ need to be propagated to versioned docs since this behavior will only be available starting in v3.3

**NOTE:** this is pending the open PR in the Stencil repo and any potential changes that will be made there.

### DIFF
--- a/docs/output-targets/01-overview.md
+++ b/docs/output-targets/01-overview.md
@@ -52,3 +52,37 @@ In the example below there are two script tags, however, only one of them will b
 As of Stencil v3, legacy browser support is deprecated, and will be removed in a future major version of Stencil.
 :::
 
+## Primary Package Output Target Validation
+
+If `validatePrimaryPackageOutputTarget: true` is set in your project's [Stencil config](../config/01-overview.md#validateprimarypackageoutputtarget) Stencil will
+attempt to validate certain fields in your `package.json` that correspond with the generated distribution code. Because Stencil can output many different formats
+from a single project, it can only validate that the `package.json` has field values that align with one of the specified output targets in your project's config.
+So, Stencil allows you to designate which output target should be used for this validation and thus which will be the default distribution when bundling you
+project. 
+
+This behavior only affects a small subset of output targets so a flag exists on the following targets that are eligible for this level of validation: `dist`, `dist-types`,
+`dist-collection`, and `dist-custom-elements`. For any of these output targets, you can configure the target to be validated as follows:
+
+```ts title='stencil.config.ts'
+import { Config } from '@stencil/core';
+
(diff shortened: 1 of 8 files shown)


rwaskiewicz (approved):

docs/output-targets/01-overview.md
```diff
@@ -52,3 +52,37 @@ In the example below there are two script tags, however, only one of them will b
...
+## Primary Package Output Target Validation
+
+If `validatePrimaryPackageOutputTarget: true` is set in your project's [Stencil config](../config/01-overview.md#validateprimarypackageoutputtarget) Stencil will
+attempt to validate certain fields in your `package.json` that correspond with the generated distribution code. Because Stencil can output many different formats
+from a single project, it can only validate that the `package.json` has field values that align with one of the specified output targets in your project's config.
+So, Stencil allows you to designate which output target should be used for this validation and thus which will be the default distribution when bundling you
+project. 
+
+This behavior only affects a small subset of output targets so a flag exists on the following targets that are eligible for this level of validation: `dist`, `dist-types`,
+`dist-collection`, and `dist-custom-elements`. For any of these output targets, you can configure the target to be validated as follows:
```
alicewriteswrongs, line 64: should we link to the pages for these output targets here, something like

```suggestion
This behavior only affects a small subset of output targets so a flag exists on the following targets that are eligible for this level of validation: [`dist`](./dist.md), `dist-types`,
`dist-collection`, and [`dist-custom-elements`](./custom-elements.md). For any of these output targets, you can configure the target to be validated as follows:
```

alicewriteswrongs (approved):
one (non-blocking) suggestion

docs/output-targets/01-overview.md
tanner-reits: I was thinking about doing this too, but felt kinda odd that we only had docs for two of the targets and wasn't sure if we wanted to draw attention to it without adding pages for the others as well 😅 

But, I suppose we can link to them for now and we can decide if we want to document the other targets as well (assuming we don't rn because they're typically just "synthetically injected" when using `dist`)

rwaskiewicz:
@tanner-reits I'll create a `stencil/v3.4-feat` branch for this to get merged into (since we're using `next` for v4 ATM)

rwaskiewicz:
https://github.com/ionic-team/stencil-site/tree/stencil/v3.4-feat created - @tanner-reits can you:
1. rebase this branch off of `stencil/v3.4-feat`
2. propagate your changes to the 3.4 directory

rwaskiewicz (approved):
TY!

PR #1128, opened 2023-05-24
docs(config): update updating guide for transformAliasedImportPaths

rwaskiewicz:
this commit updates the upgrade guide for stencil v4 to include new default values for `transformAliasedImportPaths` and `transformAliasedImportPathsInCollection`.

STENCIL-829

Core PR: https://github.com/ionic-team/stencil/pull/4418

### DIFF
--- a/docs/output-targets/dist.md
+++ b/docs/output-targets/dist.md
@@ -55,6 +55,8 @@ for more information.
 
 ### transformAliasedImportPathsInCollection
 
+*default: `true`*
+
 This option will allow [path aliases](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) defined in a project's `tsconfig.json` to be transformed into relative paths in the code output under the [collectionDir](#collectiondir) subdirectory for this output target. This does not affect imports for external packages.
 
 An example of path transformation could look something like:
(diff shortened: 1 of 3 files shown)


docs/output-targets/dist.md
```diff
@@ -67,10 +69,8 @@ import * as utils from '@utils';
 import * as utils from '../path/to/utils';
 ```
 
-This flag defaults to `false` when omitted from a Stencil configuration file.
-
 :::tip
-If using the `dist-collection` output target directly, the same result can be achieved using the `transformAliasedImportPaths` flag on the target's config.
+If using the `dist-collection` output target directly, the same result can be achieved using the [`transformAliasedImportPaths`](../config/01-overview.md#transformaliasedimportpaths) flag on the target's config.
```
tanner-reits, line 73: This link points to the top-level config flag not the one for the `dist-collection` target specifically.

tanner-reits (commented):
What are your thoughts on removing the flags specifically for `dist` and `dist-collection` and having all of the transform behavior driven by the top-level config flag? I think it's confusing having them at different levels

docs/output-targets/dist.md
rwaskiewicz: 👍 4090dda43

rwaskiewicz:
> What are your thoughts on removing the flags specifically for dist and dist-collection and having all of the transform behavior driven by the top-level config flag? I think it's confusing having them at different levels

I think it's probably the _right_ move, but I think we're a little too close to the release to change this right now. I think enabling them by default will mitigate this a little bit (it's just "on"), and something I'd be interested in coalesing for v5

tanner-reits (approved):

alicewriteswrongs (approved):

PR #1142, opened 2023-06-12
chore(repo): add initial codeowners file

rwaskiewicz:
this commit adds a codeowners file that marks the stencil team as 'owners' of all files in the repo. the intent here is _not_ to claim the site fully as our own nor box anyone out. the intent of this commit is for new pull requests to the repo to assign members of the stencil team, via codeowners, to new prs. in the future, we'll have conversations with the web team about how we can better set this up to bring them into the fold; today, we "use our best judgement" (as do they) when creating a pr/determining if we should add them to the pr to review. that'll continue as it is today until we can have those conversations


rwaskiewicz:
CC @jaredcbaum - I'd like this to get better fleshed out in the near-ish future to get your team into the fold. I need to do a little reading to shore up my codeowners knowledge and reach out to you to figure out if we can programmatically get your team on reviews. until then, I don't think much changes from your point of view (see PR summary), but wanted to give you a heads up RE this change

jaredcbaum (approved):

alicewriteswrongs (approved):

PR #1148, opened 2023-06-15
docs(angular): app initializer must be a multi-provider

sean-perkins:
Resolves: https://github.com/ionic-team/stencil-ds-output-targets/issues/354

The `APP_INITIALIZER` provider must be a multi-provider. This is a new run-time check in Angular: https://github.com/angular/angular/commit/bb6a3e849eca4487980c862b7a7e29c000a56f05

### DIFF
--- a/versioned_docs/version-v2/framework-integration/angular.md
+++ b/versioned_docs/version-v2/framework-integration/angular.md
@@ -220,9 +220,8 @@ import { defineCustomElements } from 'stencil-library/loader';
   providers: [
     {
       provide: APP_INITIALIZER,
-      useFactory: () => {
-        return defineCustomElements();
-      },
+      useFactory: () => defineCustomElements,
+      multi: true
     },
   ]
 })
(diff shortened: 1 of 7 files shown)


versioned_docs/version-v2/framework-integration/angular.md
```diff
@@ -220,9 +220,8 @@ import { defineCustomElements } from 'stencil-library/loader';
   providers: [
     {
       provide: APP_INITIALIZER,
-      useFactory: () => {
-        return defineCustomElements();
-      },
+      useFactory: () => defineCustomElements,
```
rwaskiewicz, line 223: Double checking for my own understanding, why do we return `defineCustomElements` instead of calling it now? Does Angular looks through each of these providers and invoke the return value of `useFactory`?

tanner-reits (approved):

versioned_docs/version-v2/framework-integration/angular.md
sean-perkins: Calling it was unnecessary before. Angular will invoke the function itself. They show this usage on their docs: https://angular.io/guide/dependency-injection-providers#factory-providers-usefactory (also confirmed locally in the example app for the output targets repo that I'm setting up). 

Also had Tanner test in a stackblitz 🙂

rwaskiewicz (approved):

PR #1159, opened 2023-06-23
consolidate v3 docs down to a single version

alicewriteswrongs:
This consolidates the docs we're currently maintaining for v3 down a single version in preparation for the v4 release.


vercel.json
```diff
@@ -6,6 +6,11 @@
       "destination": "/docs/introduction",
       "statusCode": 301
     },
+    {
+      "source": "/docs/:version(v3\\.[0-4])/:path*",
+      "destination": "/docs/:path*",
+      "statusCode": 301
+    },
```
alicewriteswrongs, line 13: So Vercel's documentation leaves a bit to be desired but based on reading the examples [here](https://vercel.com/docs/concepts/projects/project-configuration#redirects) and looking at the other redirects in this file I _believe_ that this will take care of redirecting any request to `/docs/$ANY_V3_VERSION/:path` to `/docs/:path`, so like if you try to view `/docs/v3.2/component-lifecycle` it will redirect to `/docs/component-lifecycle` which will show the user whatever the latest version is.

There's a _slight_ catch-22 here where until we publish v4 docs the latest version, which is rendered at `docs/` will be the consolidated `v3` version, but once we create a `v4` version (based on what's `next` right now) then `docs/:path` will point to v4 documentation.

So if we land this PR before we create the v4 docs version then we should just circle back to change this redirect to have the destination `/docs/v3/:path*` instead. If we create the v4 docs first then we should just make that change before this merges.

alicewriteswrongs (commented):
pointing out one thing

tanner-reits (approved):
Deployment looks good. I say merge this and then we'll update the redirect when we cut the v4.0 docs

rwaskiewicz (approved):
Agreed - I say let's merge this now and we'll take care of the redirect

PR #1182, opened 2023-07-20
docs(vue): update example vue library structure

verena-ifx:
The documentation is confusing when you scroll further down and you dont actually have a /src/ folder anywhere, only /lib/ and the index.ts and plugin.ts file under it.

### DIFF
--- a/docs/framework-integration/vue.md
+++ b/docs/framework-integration/vue.md
@@ -27,8 +27,8 @@ An example project set-up may look similar to:
 top-most-directory/
 └── packages/
     ├── vue-library/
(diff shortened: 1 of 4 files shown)


sean-perkins (approved):
Hello 👋 thanks for submitting this adjustment! 

I agree, this is confusing. We recommend the proxies be generated to this path:
> proxiesFile: '../vue-library/lib/components.ts',

but, that doesn't align with the example project structure we demonstrate. I'll ask someone from the team to review.

tanner-reits (approved):

PR #1191, opened 2023-07-28
docs(testing): add recipe for `setViewport`

tanner-reits:
Updates testing documentation with a recipe for `setViewport` to emulate device displays within e2e tests.

### DIFF
--- a/docs/testing/e2e-testing.md
+++ b/docs/testing/e2e-testing.md
@@ -22,9 +22,9 @@ Most methods are async and return Promises. Use `async` and `await` to declutter
 
 `E2EPage` is a wrapper utility to Puppeteer to simplify writing tests. Some helpful methods on `E2EPage` include:
 
-- `emulate(device: Device)`: Updates the page to emulate a device. This is helpful for testing a component's behavior in different orientations and viewport sizes.
 - `find(selector: string)`: Find an element that matches the selector. Similar to `document.querySelector`.
 - `setContent(html: string)`: Sets the content of a page. This is where you would include the markup of the component under test.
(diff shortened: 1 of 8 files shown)


docs/testing/e2e-testing.md
```diff
@@ -201,6 +202,21 @@ For non-shadow content:
     });
 ```
 
+### Emulate a display
```
alicewriteswrongs, line 205: Should we have a little copy here to kick things off? Just something like

> If you need to test how a component behaves with a particular viewport you can set the viewport width and height like so:

versioned_docs/version-v2/testing/config.md
```diff
@@ -72,8 +72,10 @@ export interface TestingConfig extends JestConfig {
   browserSlowMo?: number;
 
   /**
-   * Array of browser emulations to be using during e2e tests. A full e2e
+   * Array of browser emulations to be used during _screenshot_ tests. A full screenshot
    * test is ran for each emulation.
```
alicewriteswrongs, line 76:
```suggestion
   * test is run for each emulation.
```

while we're in here

alicewriteswrongs (changes requested):
looks good! just two small things

docs/testing/e2e-testing.md
rwaskiewicz: Agreed ^

rwaskiewicz (approved):
LGTM, I think @alicewriteswrongs called everything out that that we should tweak

alicewriteswrongs (approved):

PR #1192, opened 2023-07-28
Update ember.md to lean on native JS docs.

NullVoxPopuli:
Ember is compatible with all of native JS, so we can recommend folks use that.


rwaskiewicz (approved):
Thanks for this, appreciate it!

I made a few minor copy edits for Ionic products (proper casing them) and added some Docusaurus-specific sugar. I also propagated the changes to our `docs/` directory, which is the source of files when we cut a new release version of the doc site.

I'm going to ask another member of the Stencil team to give this a look, but I think we can merge this soon!

alicewriteswrongs (approved):

PR #1193, opened 2023-07-31
add spellchecking for Markdown to CI

alicewriteswrongs:
This sets up spellchecking for our Markdown files using [cspell](https://cspell.org/).

There's a new `package.json` script under the name `spellcheck` which runs the spellchecking, so you can easily do it locally before pushing to gh.


cspell-wordlist.txt
```diff
@@ -0,0 +1,31 @@
+!PostCss
```
alicewriteswrongs, line 1: This is normally stylized as `PostCSS` but we had `PostCss` several places in the codebase, so I'm disallowing that here with the `!`

See <https://cspell.org/docs/dictionaries-custom/#words-list-syntax>

cspell.json
```diff
@@ -0,0 +1,14 @@
...
+  "dictionaryDefinitions": [
+    {
+      "name": "custom-words",
+      "path": "./cspell-wordlist.txt",
+      "addWords": true
+    }
+  ],
+  "ignoreRegExpList": [
+    "/(```+)[\\s\\S]+?\\1/g",
+    "/`[a-zA-Z-_]+`/g"
```
alicewriteswrongs, line 12: the first one here ignores multi-line code blocks, the second inline code blocks

cspell-wordlist.txt
alicewriteswrongs: (we could do this with other things too in the future, like making sure we don't have `Nodejs` instead of `NodeJS` or something, just wanted to have one example of it in the first PR here)

.github/workflows/check-markdown.yml
```diff
@@ -22,3 +22,20 @@ jobs:
 
     - name: check internal markdown links
       run: npm run check-links
+
+  markdown-spelling-check:
```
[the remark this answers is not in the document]
rwaskiewicz: 👍 Do we want to gate PRs on this? My thought is yes - if so, we'll have to update the repo settings for this branch to do so after this merges

cspell.json
rwaskiewicz: Oh JSON, why won't you let us add comments to things 🙄 

I don't think it's worth converting this to YAML or CJS, but do I wish they supported JSON5

rwaskiewicz (approved):
Just one question

.github/workflows/check-markdown.yml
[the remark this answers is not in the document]
alicewriteswrongs: sure I think we could

tanner-reits (approved):
Handy little addition. Nice!

PR #1196, opened 2023-08-02
docs(config): add docs for new `experimentalSlotFixes` config value

alicewriteswrongs:
A small documentation PR just adding a bit of explanation of what this field is for.

### DIFF
--- a/docs/config/extras.md
+++ b/docs/config/extras.md
@@ -87,3 +87,23 @@ It is possible to assign data to the actual `<script>` element's `data-opts` pro
 ### slotChildNodesFix
 
 For browsers that do not support shadow dom (IE11 and Edge 18 and below), slot is polyfilled to simulate the same behavior. However, the host element's `childNodes` and `children` getters are not patched to only show the child nodes and elements of the default slot. Defaults to `false`.
+
+### experimentalSlotFixes
+
+This option enables all current and future slot-related fixes. When enabled it
+will enable the following options, overriding their values if they are
+specified separately:
+
+- [`slotChildNodesFix`](#slotchildnodesfix)
+- [`scopedSlotTextContentFix`](#scopedslottextcontentfix).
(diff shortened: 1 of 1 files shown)


rwaskiewicz (changes requested):
I think for this field, we're going to want to be pretty descriptive about what this field is, how it's being used, and what folks should expect if they enable it now (and in the future). At the moment, this is going to be one of (if not the) primary source of information for this flag. With that in mind, I think we'd want something to the effect of:

> A flag introduced in Stencil to enable new slot-related fixes to the Stencil runtime.
> When enabled, this flag will enable the following flags:
> - [list}
>
> This flag also contains fixes for the following issues that do not have a configuration flag associated with them:
> - [List of issues]
>
> Fixes to the runtime that are a part of this flag will be added over the course of Stencil v4, with the intent of making these the default behavior in Stencil v5.
>  New fixes added to this flag are not subject to Stencil's [semantic versioning policy](../reference/versioning).


The "This flag also contains fixes for the following issues" part would have to come later though, since we don't have any ATM.


But that's all just an example of what I'm thinking of in terms of things we should tell folks about. Thoughts?

tanner-reits:
Agree this should be pretty descriptive. Might be good to also add a sentence like "This option will override any values specified in the `extras` config section for slot related fixes/polyfills". This way, users understand that this takes precedence.

tanner-reits (commented):
Agree this should be pretty descriptive. Might be good to also add a sentence like "This option will override any values specified in the `extras` config section for slot related fixes/polyfills". This way, users understand that this takes precedence.

alicewriteswrongs:
@rwaskiewicz makes sense, I just updated the copy in line with your suggestions

rwaskiewicz (approved):

tanner-reits (approved):

PR #1200, opened 2023-08-07
docs(support): add puppeteer v21 to support table

rwaskiewicz:
this commit updates stencil's support docs to include puppeteer v21

Screenshot 2023-08-09 at 5 34 07 PM


docs/reference/support-policy.md
```diff
@@ -139,7 +139,7 @@ compatibility tables to describe the interoperability requirements of these piec
 
 #### Puppeteer
 
-| Stencil Version | Puppeteer v5-9 | Puppeteer v10 | Puppeteer v11-19 | Puppeteer v20 |
+| Stencil Version | Puppeteer v5-9 | Puppeteer v10 | Puppeteer v11-20 | Puppeteer v21 |
```
alicewriteswrongs, line 142: the columns for 11-20 and 21 have the same emojis in them - should we consider just consolidating to a single column?

alicewriteswrongs (changes requested):
Just one Q 🤔

docs/reference/support-policy.md
rwaskiewicz: yah - i'm ok with that 👍 updated

alicewriteswrongs (approved):

tanner-reits (approved):
Should we update v3 docs as well since, according to the table, Puppeteer v21 works with Stencil v3?

rwaskiewicz:
@tanner-reits done in https://github.com/ionic-team/stencil-site/pull/1200/commits/65b4aaeefa72b9da52d4a3457f43fd9f74526423 - I do wonder what our policy long term for things like this should be, whether to let these docs lie as-is or continue to update them. For now, since this is such a small change, seemed easier to update them so that they're aligned w/v4. Can you PTAL?

PR #1201, opened 2023-08-08
chore(spelling): add spellcheck for the root-level .md files

alicewriteswrongs:
This adds spell-checking for the root-level .md files (which is just `readme.md` and `CONTRIBUTING.md`). Adding this didn't find any spelling errors! 🎉 However, it did require adding a new ignore regex to the cspell config to ignore GitHub emoji shortcodes (which are colon-delimited) and adding 'Ionicons' to the wordlist.


cspell.json
```diff
@@ -9,6 +9,7 @@
   ],
   "ignoreRegExpList": [
     "/(```+)[\\s\\S]+?\\1/g",
-    "/`[a-zA-Z-_]+`/g"
+    "/`[a-zA-Z-_]+`/g",
+    "/:[a-z_]+:/g"
```
alicewriteswrongs, line 13: I believe this works for emoji shortcodes but after an (admittedly brief) bit of research I wasn't able to find a formal description of what emoji shortcodes can be. However! If you try `:TADA:` it doesn't work (you get :TADA: instead of :tada:) and some (like :heart_eyes:) have an underscore, so I think this should work. [Hooray for regexes](https://xkcd.com/1171/)!

rwaskiewicz (approved):

tanner-reits (approved):

PR #1202, opened 2023-08-09
fix: package names

mfranzke:
Those packages are actually organized in the `@stencil` NPM organization, but those are currently written as package names with the organization included in the name.

- [`@stencil-angular-output-target`](https://www.npmjs.com/package/@stencil-angular-output-target) doesn't exist and should be [`@stencil/angular-output-target`](https://www.npmjs.com/package/@stencil/angular-output-target)
- [`@stencil-react-output-target`](https://www.npmjs.com/package/@stencil-react-output-target) doesn't exist and should be [`@stencil/react-output-target`](https://www.npmjs.com/package/@stencil/react-output-target)
- [`@stencil-vue-output-target`](https://www.npmjs.com/package/@stencil-vue-output-target) doesn't exist and should be [`@stencil/vue-output-target`](https://www.npmjs.com/package/@stencil/vue-output-target)


rwaskiewicz (approved):
Great catch! Thanks!

PR #1224, opened 2023-09-06
docs(reactive-data): add section for watching native attributes

tanner-reits:
Adds a section to the `@Watch()` decorator docs for use with native HTML attributes. Also does a bit of re-wording to a few sentences.

### DIFF
--- a/docs/components/reactive-data.md
+++ b/docs/components/reactive-data.md
@@ -83,7 +83,17 @@ onAriaLabelChange(newVal: string, oldVal: string) {
 ```
 
 :::note
-Since native attributes are not `@Prop()` or `State()` members of the Stencil component, they will not trigger re-renders when changed.
+Since native attributes are not `@Prop()` or `State()` members of the Stencil component, they will not automatically trigger a
+re-render when changed. If you wish to re-render a component in this instance, you can leverage the `forceUpdate()` method:
+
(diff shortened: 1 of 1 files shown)


docs/components/reactive-data.md
```diff
@@ -58,17 +58,34 @@ export class LoadingIndicator {
...
+
+```tsx
+@Watch('aria-label')
+onAriaLabelChange(newVal: string, oldVal: string) {
+  console.log('Label changed:', newVal, oldVal);
+}
+```
+
+:::note
+Since native attributes are not `@Prop()` or `State()` members of the Stencil component, they will not trigger re-renders when changed.
```
alicewriteswrongs, line 86: would it make sense for us to put a little recipe here for how you might enable that behavior (i.e. having something re-render when `aria-label` changes, for instance) ?

alicewriteswrongs (changes requested):
just one question

rwaskiewicz (approved):

docs/components/reactive-data.md
tanner-reits: Good idea! Added in 003ad90dbe1ce54eeb11be93c661cd059e036e1a

alicewriteswrongs (approved):

PR #1247, opened 2023-10-09
docs for form-associated custom elements in Stencil

alicewriteswrongs:
This adds documentation of:

- building FACE generally, including how to use the `AttachInternals` decorator added in https://github.com/ionic-team/stencil/pull/4784
- the `formAssociated` option added to the argument passed to the `@Component` decorator, also added in https://github.com/ionic-team/stencil/pull/4784

### DIFF
--- a/docs/components/component.md
+++ b/docs/components/component.md
@@ -114,6 +114,40 @@ In the example above, the following allows `todo-list` to display the provided a
 
 For more information on configuring assets, please see Stencil's [Assets Guide](../guides/assets.md)
 
+
+### formAssociated
+
+**Optional**
+
+**Type: `boolean`**
+
+**Default: `false`**
+
+If `true` the component will be
+[form-associated](https://html.spec.whatwg.org/dev/custom-elements.html#form-associated-custom-element),
+allowing you to take advantage of the
+[`ElementInternals`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/attachInternals)
+API to enable your Stencil component to participate in forms.
+
+A minimal form-associated Stencil component could look like this:
+
+```tsx
+import { Component } from '@stencil/core';
+
+@Component({
+  tag: 'form-associated',
+  formAssociated: true
+})
+export class FormAssociated {
+  render() {
+    return <span>form associated!</span>
+  }
+}
+```
+
+See the documentation for [form-associated components](./form-associated.md)
+for more info and examples.
+
 ### scoped
 
 **Optional**
--- a/docs/components/form-associated.md
+++ b/docs/components/form-associated.md
@@ -1,0 +1,132 @@
+---
+title: Form-Associated Components
+sidebar_label: Form-Associated Components
+description: Form-Associated Stencil Components
+slug: /form-associated
+---
+
+# Building Form-Associated Components in Stencil
+
+As of v4.5.0, Stencil has support for form-associated custom elements. This
+allows Stencil components to participate in a rich way in HTML forms,
+integrating with native browser features for validation and accessibility while
+maintaining encapsulation and control over their styling and presentation.
+
+:::caution
+Browser support for the APIs that this feature depends on is still not
+universal[^1] and the Stencil team has no plans at present to support or
+incorporate any polyfills for the browser functionality. Before you ship
+form-associated Stencil components make sure that the browsers you need to
+support have shipped the necessary APIs.
+:::
+
+## Creating a Form-Associated Component
+
+A form-associated Stencil component is one which sets the new [`formAssociated`](./component.md#formAssociated)
+option in the argument to the `@Component`
+decorator to `true`, like so:
+
+```tsx
+import { Component } from '@stencil/core';
+
+@Component({
+  tag: 'my-face',
+  formAssociated: true,
+})
+export class MyFACE {
+}
+```
+
+This element will now be marked as a form-associated custom element via the
+[`formAssociated`](https://html.spec.whatwg.org/#custom-elements-face-example)
+static property, but by itself this is not terribly useful.
+
+In order to meaningfully interact with a `<form>` element that is an ancestor
+of our custom element we'll need to get access to an
+[`ElementInternals`](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals)
+object corresponding to our element instance. Stencil provides a decorator,
+`@AttachInternals`, which does just this, allowing you to decorate a property on
+your component and bind an `ElementInternals` object to that property which you
+can then use to interact with the surrounding form.
+
+:::info
+Under the hood the `AttachInternals` decorator makes use of the very similarly
+named
+[`attachInternals`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/attachInternals)
+method on `HTMLElement` to associate your Stencil component with an ancestor
+`<form>` element. During compilation, Stencil will generate code that calls
+this method at an appropriate point in the component lifecycle for both
+[lazy](../output-targets/dist.md) and [custom
+elements](../output-targets/custom-elements.md) builds.
+:::
+
+A Stencil component using this API to implement a custom text input could look
+like this:
+
+```tsx title="src/components/custom-text-input.tsx"
+import { Component, h, AttachInternals, State } from '@stencil/core';
+
+@Component({
+  tag: 'custom-text-input',
+  shadow: true,
+  formAssociated: true
+})
+export class CustomTextInput {
+  @State() value: string;
+
+  @AttachInternals() internals: ElementInternals;
+
+  handleChange(event) {
+    this.value = event.target.value;
+    this.internals.setFormValue(event.target.value);
+  }
+
+  componentWillLoad() {
+    this.internals.setFormValue("a default value");
+  }
+
+  render() {
+    return (
+      <input
+        type="text"
+        value={this.value}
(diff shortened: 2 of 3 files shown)


docs/components/component.md
```diff
@@ -114,6 +114,36 @@ In the example above, the following allows `todo-list` to display the provided a
 
 For more information on configuring assets, please see Stencil's [Assets Guide](../guides/assets.md)
 
+
+### `formAssociated`
```
rwaskiewicz, line 118: Minor, can we remove these backticks? We don't use them on any other properties for the decorator, and they're causing some styling differences:
Screenshot 2023-10-09 at 3 14 11 PM

In the image above, the red circles are meant to show the difference in the background & font of these headers. The blue calls out the same for the sidebar

docs/components/component.md
```diff
@@ -114,6 +114,36 @@ In the example above, the following allows `todo-list` to display the provided a
...
+
+**Optional**
+
+If `true` the component will be
```
rwaskiewicz, line 122: Can we add type and default values to match the other fields here?

```suggestion
*Type: `boolean`**

**Default: `false`**

If `true` the component will be
```

docs/components/form-associated.md
```diff
@@ -0,0 +1,110 @@
+---
+title: Form-Associated Components
+sidebar_label: Form-Associated Components
+description: Form-Associated Stencil Components
+slug: /form-associated
+---
+
+# Building Form-Associated Components in Stencil
+
+As of $VERSION Stencil has experimental support for form-associated custom
```
rwaskiewicz, line 10: I feel _pretty confident_ this'll be 4.5.0. If that's not the case, that's fine -d just adding this comment so we don't forget.
```suggestion
As of 4.5.0 Stencil has experimental support for form-associated custom
```

docs/components/form-associated.md
```diff
@@ -0,0 +1,110 @@
...
+accessibility while maintaining encapsulation and control over their styling
+and presentation.
+
+:::caution
+Browser support for the APIs that this feature depends on is still relatively
+low ([~84% as of this writing](https://caniuse.com/?search=attachInternals))
+and the Stencil team has no plans at present support or incorporate any
+polyfills for the browser functionality. Accordingly, this is a 'use at your
+own risk' feature: it is up to you as a developer to ensure the browsers you
+need to support have shipped the necessary APIs.
```
rwaskiewicz, line 22: This reads a little too adversarial to me - I think we can soften this a little bit by removing the content that precedes the colon like so:
```suggestion
polyfills for the browser functionality. It is the responsibility of the developer to ensure the browsers you need to support have shipped the necessary APIs.
```

docs/components/form-associated.md
```diff
@@ -0,0 +1,110 @@
...
+:::
+
+## Creating a form-associated component
```
rwaskiewicz, line 25: Since this is a title, can we capital-case it?
```suggestion
## Creating a Form-Associated Component
```

docs/components/form-associated.md
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+```
+
+This component doesn't do a whole lot as-is, but luckily a great deal more is
+possible with the `ElementInternals` API, including [setting the element's
+validity](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals/setValidity),
+reading the validity state of the form, reading other form values, and more.
+
+### Resources
```
rwaskiewicz, line 107: Should this be a child header of 'Creating a form-associated component', or should this stand as a sibling to it?
```suggestion
## Resources
```

docs/components/form-associated.md
rwaskiewicz, line 10: Since we normally put experimental features behind flags, should we s/experimental/initial?
```suggestion
As of $VERSION Stencil has initial support for form-associated custom
```

docs/components/form-associated.md
rwaskiewicz: I'm also a little concerned about this figure (~84% as of this writing) changing over time/us not keeping it up to date. What're you're thoughts on something like:

> Browser support for the APIs that this feature depends are still evolving.
At this time, the Stencil team has no plans at present support or incorporate any
polyfills for the browser functionality. It is the responsibility of the developer to ensure the browsers you need to support have shipped the necessary APIs.

docs/components/form-associated.md
```diff
@@ -0,0 +1,110 @@
...
+and the Stencil team has no plans at present support or incorporate any
+polyfills for the browser functionality. Accordingly, this is a 'use at your
+own risk' feature: it is up to you as a developer to ensure the browsers you
+need to support have shipped the necessary APIs.
+:::
+
+## Creating a form-associated component
+
+A form-associated Stencil component is one which sets the new `formAssociated`
+option in the argument to the `@Component`
```
rwaskiewicz, line 28: Can we link this back to the `formAssociated` option on the component page?
```suggestion
A form-associated Stencil component is one which sets the new [`formAssociated`
option](./component.md#formAssociated) in the argument to the `@Component`
```

docs/components/form-associated.md
```diff
@@ -0,0 +1,110 @@
...
+        type="text"
+        value={this.value}
+        onInput={(event) => this.handleChange(event)}
+      />
+    )
+  }
+}
+```
+
+This component doesn't do a whole lot as-is, but luckily a great deal more is
```
rwaskiewicz, line 102: Can we walk folks through what it does do? We put the onus on readers here to understand what the code does, and learn a new concept - I think it could be beneficial to folks for us to explain what's occurring - similar to what we do in https://stencil-docs-git-ap-form-associated-ionic1.vercel.app/docs/next/getting-started#anatomy-of-my-component and tie it back to what's been described already.

rwaskiewicz (changes requested):
I think this is a good start! 

One larger question I have is how this would interact with our guide for forms - https://stencil-docs-git-ap-form-associated-ionic1.vercel.app/docs/next/forms. Should we update that as well?

docs/components/component.md
alicewriteswrongs: sure, although I'd maybe argue they should all have them! haha

docs/components/form-associated.md
alicewriteswrongs: I'll misspell `VERSION` on purpose so we can't merge till we update

docs/components/form-associated.md
alicewriteswrongs: Sure, although I think "as of this writing" indicates it's not going to be current forever - once it's higher (above what threshold I don't know) I guess the note should just go away entirely

docs/components/form-associated.md
alicewriteswrongs: yeah, that might be something to discuss a little bit - I think we floated this being an 'experimental' feature but I don't know if we made a solid decision. The advantage of doing so I guess is that we can change APIs more freely if we run into some problems with them but I'm not sure what exactly makes the most sense

docs/components/form-associated.md
alicewriteswrongs: yea I think I wanted two `#` here

docs/components/form-associated.md
alicewriteswrongs: yeah I think that's a good idea, it's just one method call here but why not explain it. I added an explanation of how it works, one little quirk (reading the `"name"` attribute off of the host element) and links to some MDN articles

docs/components/form-associated.md
```diff
@@ -0,0 +1,133 @@
...
+
+As of $VERISION Stencil has initial support for form-associated custom
+elements. This allows Stencil components to participate in a rich way in HTML
+forms, integrating with native browser features for validation and
+accessibility while maintaining encapsulation and control over their styling
+and presentation.
+
+:::caution
+Browser support for the APIs that this feature depends on is still not
+universal[^1] and the Stencil team has no plans at present support or
```
tanner-reits, line 18:
```suggestion
universal[^1] and the Stencil team has no plans at present to support or
```

docs/components/form-associated.md
tanner-reits: I was under the impression we were initially shipping this as experimental. The feature PR also states it as experimental

docs/components/form-associated.md
alicewriteswrongs: 🤦‍♀️

docs/components/form-associated.md
alicewriteswrongs: yeah, it feels like we've landed on something that will work and that we'd only possibly add to later, i.e. adding some nicer API around validation or something like that (just spitballing not proposing haha). like what we have now feels very close to the spec and minimal, so I wouldn't anticipate we'd need to massively rework parts of it in the future and so an actual 'experimental' label wouldn't precisely be warranted

alicewriteswrongs: To round this out, I think it makes sense to just say "external" here and to leave the `@Component` option and the decorator as-is for now

rwaskiewicz: > To round this out, I think it makes sense to just say "external" here and to leave the @Component option and the decorator as-is for now

I think I'm lost, what parts 'external'?

alicewriteswrongs: ohh `s/external/initial` 🤪

rwaskiewicz (approved):
LGTM - FWIW, you have one (non-version related) typo, but non-blocking

PR #1284, opened 2023-12-04
docs(angular): updated instructions for angular modules vs. standalone

sean-perkins:
Updates the Framework Wrapper docs for Angular to separate usage instructions based on an Angular application using modules or standalone. 

Additionally updates the markdown code block syntax to use `title` to clearly outline which files the code blocks refer to. 

Preview: https://stencil-docs-git-sp-fw-5678-ionic1.vercel.app/docs/next/angular#consumer-usage


docs/framework-integration/angular.md
```diff
@@ -309,15 +305,58 @@ export class AppModule {}
...
+import { MyComponent } from 'component-library';
+
+@Component({
+  selector: 'app-root',
+  templateUrl: './app.component.html',
+  styleUrls: ['./app.component.css'],
+  standalone: true,
+  imports: [MyComponent],
+})
+export class AppComponent {}
```
rwaskiewicz, line 349: Using this snippet, I ran into the following error message when trying to run hte local dev server for the app:
```
[ERROR] TS-992011: The component 'MyComponent' appears in 'imports', but is not standalone and cannot be imported directly. It must be imported via an NgModule. [plugin angular-compiler]

    projects/my-app/src/app/app.component.ts:9:12:
      9 │   imports: [MyComponent],
        ╵             ~~~~~~~~~~~
```

I ran through the instructions here from start to end to verify that I wasn't starting with a pre-existing reproduction locally that was in a bad state. 

Do we need to call out to folks that when they need to use 'standalone' for the [outputType](https://stencil-docs-git-sp-fw-5678-ionic1.vercel.app/docs/next/angular#outputtype) when they're configuring the Stencil OT here? I _think_ that'd lead to quite a few more changes, but I don't fully understand the error message above and wanted to double check before we went down that road

sean-perkins: Ahh yes, we need to additionally reference each usage for the different `outputType` configurations. I'll think on how we might split this up. I think separating consumer usage by type makes sense, but maybe another subsection within it for the different output types.

rwaskiewicz (approved):
This makes sense to me - @tanner-reits could you do a pass when you get a chance?

PR #1292, opened 2023-12-13
docs(events): update type conflict recommendation

tanner-reits:
Updates the verbiage around the recommendation for dealing with type conflicts in the `Event` decorator import with the addition of support for import aliasing.

Also peeked through the docs for some of our other decorators and didn't see anything needing updating there.

**Note:** This does **not** need to be propagated to other versions as this will only be available starting in v4.9

### DIFF
--- a/docs/components/events.md
+++ b/docs/components/events.md
@@ -83,9 +83,11 @@ export class TodoList {
 ```
 
 :::note
-In the case where the Stencil `Event` type conflicts with the native web `Event` type, it is suggested that the native web `Event` type be prefixed with `globalThis`:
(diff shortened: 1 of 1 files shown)


rwaskiewicz (changes requested):
Can we show both options? I think the advice here could still be beneficial for a few scenarios:
1. If/when folk find the v4.9+ docs through a Google search, when they're on v4.8 or lower. They'd be able to use the advice we have live today.
2. If some other bug/issue prevents them from using the aliases

rwaskiewicz (approved):

PR #1313, opened 2024-01-09
docs(component): add note on limitation of rendering referenced components

christian-bromann:
Based on investigations done in https://github.com/ionic-team/stencil/issues/3953 we don't recommend importing components directly and instead just use the component tags.


versioned_docs/version-v4.1/components/functional-components.md
```diff
@@ -102,3 +102,7 @@ There are a few major differences between functional components and class compon
 * are stateless.
 
 When deciding whether to use functional components, one concept to keep in mind is that often the UI of your application can be a function of its state, i. e., given the same state, it always renders the same UI. If a component has to hold state, deal with events, etc, it should probably be a class component. If a component's purpose is to simply encapsulate some markup so it can be reused across your app, it can probably be a functional component (especially if you're using a component library and thus don't need to style it).
+
+:::caution
+It is not recommended to re-export a functional component from a "barrel file" and attempt to dynamically render it in another component. This is a [known limitation](https://github.com/ionic-team/stencil/issues/5246) within Stencil. Instead, either use class components and remove the import or import the functional component directly.
```
alicewriteswrongs, line 107: I just wonder if this should be a bit stronger than a recommendation, and maybe we should more directly say that this won't work, perhaps something like:

```suggestion
Stencil does not support re-exporting a functional component from a "barrel file" and dynamically rendering it in another component. This is a [known limitation](https://github.com/ionic-team/stencil/issues/5246) within Stencil. Instead, either use class components and remove the import or import the functional component directly.
```

alicewriteswrongs (changes requested):
one question

rwaskiewicz (approved):
FWIW, I'm good with either verbiage - either works IMO

versioned_docs/version-v4.1/components/functional-components.md
christian-bromann: I like a more explicit language. Thank you!

alicewriteswrongs (changes requested):
looks great to me with the updated copy, we just want to propagate that to all of our files

christian-bromann:
@alicewriteswrongs duh' .. updated all other locations too.

rwaskiewicz (approved):

PR #1316, opened 2024-01-10
docs(testing): document how to pass in Jest CLI arguments

christian-bromann:
Based on a question we've received in https://github.com/ionic-team/stencil/discussions/5225#discussioncomment-8084146 I think it is worth to document this.

### DIFF
--- a/docs/testing/01-overview.md
+++ b/docs/testing/01-overview.md
@@ -90,13 +90,13 @@ export const config: Config = {
 
 ### Command Line Arguments
 
-While the Stencil CLI offers a certain set of command line flags to specify e.g. which types of tests to run, you also have access to all Jest options through the CLI. For example to specify a single test, you can pass in a positional argument to the command, e.g.:
+While the Stencil CLI offers a certain set of command line flags to specify e.g. which types of tests to run, you also have access to all Jest options through the CLI. For example to specify a single test, you can pass in a positional argument to Jest by adding a `--`, e.g.:
 
 ```sh
 # run a single unit test
-npx stencil test --spec src/components/my-component/my-component.spec.ts
+npx stencil test --spec -- src/components/my-component/my-component.spec.ts
 # run a single e2e test
-npx stencil test --e2e src/components/my-component/my-component.e2e.ts
+npx stencil test --e2e -- src/components/my-component/my-component.e2e.ts
 ```
 
 Next to positional arguments, Stencil also passes along [certain](https://github.com/ionic-team/stencil/blob/54d4ee252768e1d225baababee0093fdb0562b83/src/cli/config-flags.ts#L38-L85) Jest specific flags, e.g.:
--- a/docs/testing/01-overview.md
+++ b/docs/testing/01-overview.md
@@ -88,6 +88,19 @@ export const config: Config = {
 };
 ```
 
+### Command Line Arguments
+
+While the Stencil CLI offers a certain set of command line flags to specify e.g. which types of tests to run, you also have access to all Jest options through the CLI by separating them with `--`. For example to specify a single test, you can call:
+
+```sh
+# run a single unit test
+npx stencil test --spec -- src/components/my-component/my-component.spec.ts
+# run a single e2e test
+npx stencil test --e2e -- src/components/my-component/my-component.e2e.ts
+```
+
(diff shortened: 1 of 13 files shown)


rwaskiewicz (approved):
@alicewriteswrongs are there any cases that you know of where Stencil overrides Jest configuration flags? IIRC, there is some overlap between certain options, but I'm unsure if they're worth calling out here due to functional differences

alicewriteswrongs:
The Jest config is built here: https://github.com/ionic-team/stencil/blob/main/src/testing/jest/jest-29/jest-config.ts#L12

It will use both `knownArgs` (which are arguments that Stencil supports and recognized) as well as all positional arguments passed at the comment line (e.g. a filename pattern as shown in the change made in this PR) in order to generate a Jest config.

I believe that given the way the command-line arguments are marshalled before they're passed to Jest any Jest arguments that are passed should be passed through without Stencil editorializing them. Some of those Jest arguments may just also have a side-effect of setting a particular Stencil configuration value

alicewriteswrongs:
also strictly speaking the `--` is not for passing in _jest_ CLI flags, but rather positional arguments like the filename pattern. We support all of hte Jest CLI flags as Stencil CLI options, those are enumerated here: https://github.com/ionic-team/stencil/blob/54d4ee252768e1d225baababee0093fdb0562b83/src/cli/config-flags.ts#L38-L85

Our parser will set all of those on the `ConfigFlags` object that we create during CLI argument parsing, but only the really Stencil-specific options will have an effect on the Stencil config for a project when it's building

christian-bromann:
@alicewriteswrongs thanks for the insights. I updated the patch to reflect what you are describing, does it make sense? Do you see anything missing we should mention?

docs/testing/01-overview.md
```diff
@@ -88,6 +88,26 @@ export const config: Config = {
 };
 ```
 
+### Command Line Arguments
+
+While the Stencil CLI offers a certain set of command line flags to specify e.g. which types of tests to run, you also have access to all Jest options through the CLI. For example to specify a single test, you can pass in a positional argument to the command, e.g.:
+
+```sh
+# run a single unit test
+npx stencil test --spec src/components/my-component/my-component.spec.ts
```
rwaskiewicz, line 97: I think you still need the argument separator (`--`) here. On a sample project (`npm init stencil@latest component link-test`) that has one additional component (`npm run generate`, enter 'link-me' for the test name), this runs all of the tests. If I run this command with `--` to denote the file path as a positional arg, it runs just those tests:
```
➜  link-test git:(main) ✗ npx stencil test --spec -- src/components/my-component/my-component.spec.ts
[34:34.3]  @stencil/core
[34:34.4]  v4.9.1 🍬
[34:34.5]  testing spec files
[34:34.5]  jest args: --spec --max-workers=8 -- src/components/my-component/my-component.spec.ts
 PASS  src/components/my-component/my-component.spec.ts
  my-component
    ✓ renders (8 ms)
    ✓ renders with values (2 ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        0.241 s, estimated 1 s
Ran all test suites matching /src\/components\/my-component\/my-component.spec.ts/i.

➜  link-test git:(main) ✗ npx stencil test --spec src/components/my-component/my-component.spec.ts
[34:41.9]  @stencil/core
[34:42.1]  v4.9.1 🍬
[34:42.1]  testing spec files
[34:42.1]  jest args: --spec --max-workers=8
 PASS  src/components/my-component/my-component.spec.ts
 PASS  src/components/link-me/test/link-me.spec.tsx
 PASS  src/utils/utils.spec.ts

Test Suites: 3 passed, 3 of 5 total
Tests:       7 passed, 7 total
Snapshots:   0 total
Time:        0.341 s, estimated 1 s
Ran all test suites.
```

rwaskiewicz (changes requested):

docs/testing/01-overview.md
christian-bromann: > I think you still need the argument separator (`--`) here.

Sorry I was meaning to remove `--spec` here to showcase positional arguments

christian-bromann: @rwaskiewicz thanks, I forgot about that when restructuring the docs. Pushed an update.

PR #1329, opened 2024-01-22
chore(): set `onBrokenAnchors` to 'throw' and fix all errors

alicewriteswrongs:
Docusaurus has an option called [`onBrokenAnchors`](https://docusaurus.io/docs/api/docusaurus-config#onBrokenAnchors) which, as the name suggests, controls how it should handle anchors which do not resolve to a heading in the page to which they link.

This sets the option to `throw`, so that the build will fail if we have a broken anchor on a link.

It turns out we have quite a few! So this PR also fixes them.

A neat think is that it will even check the `to` prop of [Docusaurus' built-in `<Link>` component](https://docusaurus.io/docs/docusaurus-core#link) in a React component, which caught a typo on our lifecycle methods chart. Neat!


tanner-reits (approved):
Good to get those fixed!

christian-bromann (approved):

PR #1330, opened 2024-01-24
docs(jsx): add explanation of automatic key insertion

alicewriteswrongs:
Add some documentation to our page on JSX which explains the changes made in https://github.com/ionic-team/stencil/pull/5143

### DIFF
--- a/docs/components/templating-and-jsx.md
+++ b/docs/components/templating-and-jsx.md
@@ -248,6 +248,46 @@ ensure that each child has a key which does not change and which is unique
 among all its siblings.
 :::
 
+### Automatic Key Insertion
+
+During compilation Stencil will automatically add key attributes to any JSX
+nodes in your component's render method which are not nested within curly
+braces. This allows Stencil’s runtime to accurately reconcile children when
+their order changes or when a child is conditionally rendered.
+
+For instance, consider a render method looking something like this:
+
+```tsx
+  render() {
+    return (
+      <div>
+        { this.disabled && <div id="no-key">no key!</div> }
+        <div id="slot-wrapper">
+          <slot/>
+        </div>
+      </div>
+    );
+  }
+```
+
+While it might seem like adding a key attribute to the `#slot-wrapper` div
+could help ensure that elements will be matched up correctly when the component
+re-renders, this is actually superfluous because Stencil will automatically add
+a key to that element when it compiles your component.
+
+:::note
+The Stencil compiler can only safely perform automatic key insertion in certain
+scenarios where there is no danger of the keys accidentally causing elements to
+be considered different when they should be treated the same (or vice versa).
+
+In particular, the compiler will not automatically insert `key` attributes if a
+component's `render` method has more than one `return` statement or if it
+returns a [conditional
+expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator).
+Additionally, the compiler will not add key attributes to any JSX which is
+found within curly braces (`{ }`).
+:::
+
 ## Handling User Input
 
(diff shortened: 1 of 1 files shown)


christian-bromann (commented):
When reading the new section I asked myself how the given example is different from the one just above it:
```tsx
render() {
  return (
    <div>
      {this.todos.map((todo) => (
        <div key={todo.uid}>
          <div>{todo.taskName}</div>
        </div>
      ))}
    </div>
  )
}
```
Is this example outdated now since Stencil adds keys automatically?

alicewriteswrongs:
hmm probably I need to explain better, I'll re-word a bit.

In short it's not outdated, because we decided in the implementation of https://github.com/ionic-team/stencil/pull/5143 not to insert keys into JSX nodes which are within curly braces (a JSX expression) because there are just so many possibly things that _could_ be going on there that it's very hard to know when it would be safe to add keys to elements without breaking the user's expectation for how their component should render.

An example would be a usage of `Array.prototype.map` like in that example. If the user had a component with something like:

```tsx
render() {
  return (
    <div>
      {this.todos.map((todo) => (
        <div>
          <div>{todo.taskName}</div>
        </div>
      ))}
    </div>
  )
}
```

and we inserted keys automatically the transformed code would look something like this:

```tsx
render() {
  return (
    <div>
      {this.todos.map((todo) => (
        <div key="a-key">
          <div>{todo.taskName}</div>
        </div>
      ))}
    </div>
  )
}
```

this is because the transformation is done at compilation-time and on the basis of the syntax tree nodes that the transformer finds, not at runtime based on the actual data found in `this.todos`. This would create a problem where the keys would make all the children look identical. So for that reason we need to not insert keys in three situations:

1. within a JSX expression (i.e. the curly braces in `<div>{ foo }</div>`
2. when a `render` method has more than one return
3. when the first child of the `return` keyword in a `render` method is a conditional expression (aka ternary)

the 2nd and 3rd are necessary for us to avoid because in each we can't assume that the user's intent is for the two (or more) different branches in 'return behavior' are supposed to be matched up, and inserting keys in that situation would change the rendering behavior for a lot of cases.

Anywhoo - hopefully that helps explain a bit of how this works, I'll update to clarify more what is meant by the transformation not happening within a JSX expression

christian-bromann:
Thanks @alicewriteswrongs , that clears things up!

alicewriteswrongs:
@christian-bromann I updated this a bit, lmk what you think!

christian-bromann (approved):
LGTM 👍

docs/components/templating-and-jsx.md
```diff
@@ -248,6 +248,48 @@ ensure that each child has a key which does not change and which is unique
...
+
+For instance, consider a render method looking something like this:
+
+```tsx
+  render() {
+    return (
+      <div>
+        {this.disabled && (
+          <div />
+        )}
```
rwaskiewicz, line 266: I think we could strengthen this section a little bit by making it clearer that it's in curly braces whereas the "slot-wrapper" is not.

I think we could accomplish that by:
1. Adding whitespace (newlines) after the opening `{` and before the closing `}`
2. Adding a text node and/or id to the div that helps draw attention to it being in curlies:

```suggestion
        {
          this.disabled && <div id="in-curly-braces">Won't receive automatic key insertion</div>
        }
```

☝️ That might not be the _best_ in terms of presentation (particularly may be too wide for the viewport), but is the general idea. Thoughts?

rwaskiewicz (approved):
LGTM - one minor suggestion that's non-blocking

docs/components/templating-and-jsx.md
alicewriteswrongs: mmm I think I'll go for something like

```tsx
  { this.disabled && <div id="no-key">no key!</div> }
```

just to make sure it's narrow enough

PR #1338, opened 2024-01-29
docs(framework-integrations): add note about consuming custom elements

tanner-reits:
This adds an admonition to each of the framework wrapper output targets about consuming the output of `dist-custom-elements`. 

Since Stencil v4, we had to introduce [`customElementExportBehavior`](https://stenciljs.com/docs/custom-elements#customelementsexportbehavior) and the build no-longer automatically has the output of `single-export-module` behavior. Unfortunately, our framework wrappers rely on this behavior when generating the proxies, so devs need to be aware to explicitly set this option.

Note: I will propagate this changes to all v4 versions once approved

### DIFF
--- a/docs/framework-integration/angular.md
+++ b/docs/framework-integration/angular.md
@@ -594,6 +594,12 @@ Specifies the type of output to be generated. It can take one of the following v
 
 Both `scam` and `standalone` options are compatible with the `dist-custom-elements` output.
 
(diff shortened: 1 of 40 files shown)


christian-bromann (approved):
👍

rwaskiewicz (approved):

PR #1342, opened 2024-02-03
docs(style-modes): document style modes

### LINKED ISSUE #1214: Missing content in styleUrls docs
1. Visit https://stenciljs.com/docs/component#styleurls
2. See garbled content in the details section:

```
Alternatively, an object can be provided that maps a named "mode" to one or more stylesheets. This

By out of the box, Stencil will only process CSS files (ending with .css). 
```

christian-bromann:
fixes #1214
refs `STENCIL-1042`

This patch adds some documentation for style modes. I decided to add a section for this to the `Styling` page and link to it from the component as well as API docs.


docs/components/styling.md
```diff
@@ -233,6 +233,66 @@ By specifying "inner-text" as the value of the `exportparts` attribute, elements
...
+
+const defaultMode = 'md'; // Default to Material Design
+setMode((el) => el.getAttribute('colorMode') || defaultMode);
+```
+
+Style modes can only be set once at the beginning of the component lifecycle.
+
+### Important Considerations
+
+- __Initialization:__ Style modes must be defined at the start of the component lifecycle and cannot be changed thereafter.
```
tanner-reits, line 283: I think we should expand on this a bit and clarify what "beginning of the component lifecycle" means. Like does this need to happen before bootstrap? Can it be set in any lifecycle callbacks?

docs/components/styling.md
```diff
@@ -233,6 +233,66 @@ By specifying "inner-text" as the value of the `exportparts` attribute, elements
...
+- __Usage Requirement:__ A style mode must be set to ensure the component loads with styles. Without specifying a style mode, the component will not apply any styles.
+- __Querying Style Mode:__ To check the current style mode and e.g. provide different functionality based on the mode, use the `getMode` function:
+
+```ts
+import { setMode, getMode } from '@stencil/core';
```
tanner-reits, line 288:
```suggestion
import { getMode } from '@stencil/core';
```

tanner-reits (changes requested):
Small suggestion

docs/components/styling.md
christian-bromann: Updated the wording to:

```
- __Initialization:__ Style modes must be defined before the components gets rendered into the DOM and can't be changed afterwards. Updating the style mode requires manually re-rendering of all components.
```

If you update a style mode you have to re-render the whole component, meaning: remove it from the DOM and add it back in. Maybe there is a helper method that could do that, something like `requestUpdate` in Lit? Stencil has to re-initialize the component including its styles.

I may suggest two options here:
- don't recommend to ever update the mode once initially set
- ingest a bug ticket into our roadmap to investigate mode updates during runtime

docs/components/styling.md
```diff
@@ -233,6 +233,66 @@ By specifying "inner-text" as the value of the `exportparts` attribute, elements
...
 
+## Style Modes
+
+Component Style Modes enable you to create versatile designs for your components by utilizing different styling configurations. This is achieved by assigning the styleUrls property of a component to a collection of style mode names, each linked to their respective CSS files.
+
+### Example: Styling a Button Component
+
+Consider a basic button component that supports both iOS and Material Design aesthetics:
+
+```tsx
```
rwaskiewicz, line 244: Can we add a title to this code block?
```suggestion
```tsx title="Using style modes to style a component"
```

docs/components/styling.md
```diff
@@ -233,6 +233,66 @@ By specifying "inner-text" as the value of the `exportparts` attribute, elements
...
+  styleUrls: {
+    md: './simple-button.md.css', // styles for Material Design
+    ios: './simple-button.ios.css' // styles for iOS
+  },
+})
+export class SimpleButton {
+  // ...
+}
+```
+
```
rwaskiewicz, line 256: Can we add a sentence or two about the content of `styleUrls` above? I think there might be room to explain it a bit more/describe what's happening here. What do you think about:
```suggestion

In the example above, two different modes are declared.
One mode is named `md` (for 'Material Design') and refers back to a Material Design-specific stylesheet.
Likewise, the other is named `ios` (for iOS) and references a different stylesheet for iOS-like styling.
Both stylesheets are relative paths to the file that declares the component.
While we have chosen short names in the above example, there's no limitation to the keys used in the `styleUrls` object.

```

docs/components/styling.md
```diff
@@ -233,6 +233,66 @@ By specifying "inner-text" as the value of the `exportparts` attribute, elements
...
+const defaultMode = 'md'; // Default to Material Design
+setMode((el) => el.getAttribute('colorMode') || defaultMode);
+```
+
+Style modes can only be set once at the beginning of the component lifecycle.
+
+### Important Considerations
+
+- __Initialization:__ Style modes must be defined before the components gets rendered into the DOM and can't be changed afterwards. Updating the style mode requires manually re-rendering of all components.
+- __Usage Requirement:__ A style mode must be set to ensure the component loads with styles. Without specifying a style mode, the component will not apply any styles.
```
rwaskiewicz, line 284: Can we add some additional information regarding what happens if:
- We mistype the mode (e.g. `setMode('oos')`)
- A component doesn't have any style that matches the arg passed to `setMode`, but other components do?

rwaskiewicz (changes requested):

rwaskiewicz:
Ah, could we also add these docs to the v4.12 docs page as well?

docs/components/styling.md
christian-bromann: LGTM 👍

docs/components/styling.md
christian-bromann: Good catch 👌 

> We mistype the mode (e.g. setMode('oos'))

I will just point out that a user should ensure that set modes should get validated otherwise components remain unstyled.

> A component doesn't have any style that matches the arg passed to setMode, but other components do?

A style mode assigned to a component that doesn't defines that style mode will remain unstyled.

christian-bromann: I wonder if there is a way to get all defined style modes 🤔 if components define different style modes there might use cases where one would like to know which are supported by a component.

rwaskiewicz: There isn't for users today, but I don't think that'd be too big of a lift - we have that style [mode resolution list](https://github.com/ionic-team/stencil/blob/80ef14361613706d38ad6799a03d3e5f1e1e6cbe/src/runtime/mode.ts#L9) _and_ this should be pretty easy for us to derive for specific components at compile time for more details (e.g. you wanted to see if any components are missing a mode of some particular name).

The Framework team/Mike H have asked us about a typed version of `getMode`/`setMode` in the past as well - one where you can specify _somewhere_ the allowable types/strings on those methods to prevent typos in the first place. I don't think we have something in the backlog, but could be something worth asking in our internal ask_framework channel.

christian-bromann: Definitely worth adding IMHO. Added the following limitation to https://github.com/ionic-team/stencil-site/pull/1342/commits/a329dd5fd6909686b27b262bf580067f673e9901:
```
- __Input Validation:__ Verify a style mode is supported by a component you are setting it for. Setting an un-supported style mode keeps the component unstyled.
```

christian-bromann:
> Ah, could we also add these docs to the v4.12 docs page as well?

Done

docs/components/styling.md
rwaskiewicz: 👍 Can you do me a favor and add a ticket to the backlog for us to look into?

rwaskiewicz (approved):

versioned_docs/version-v2/components/styling.md
```diff
@@ -233,6 +233,70 @@ By specifying "inner-text" as the value of the `exportparts` attribute, elements
...
+```
+
+You can conditionally set the style mode based on the `colorMode` property:
+
+```ts
+import { setMode } from '@stencil/core';
+
+const defaultMode = 'md'; // Default to Material Design
+setMode((el) => el.getAttribute('colorMode') || defaultMode);
+```
```
rwaskiewicz, line 279: Double checking something - can you point me to where we check `colorMode` in Stencil?

christian-bromann: We don't, this is just an example. I learned today that the Framework team uses `mode` as property name for essentially the same. It is not required to use any specific property name here.

christian-bromann: Let me update to just use `mode` here as `color` can be confusing.

docs/components/styling.md
christian-bromann: Ingested into backlog as `STENCIL-1124`

rwaskiewicz (approved):

versioned_docs/version-v2/components/styling.md
rwaskiewicz: Thanks! LGTM

tanner-reits (approved):

PR #1344, opened 2024-02-05
docs(testing): re-organize testing up to allow adding more frameworks

christian-bromann:
Based on an internal design doc we'ld like to re-organise the Stencil docs around testing so we can add more frameworks moving on.


docs/testing/stencil-testrunner/02-config.md
```diff
@@ -0,0 +1,20 @@
...
+The `testing` config setting in `stencil.config.ts` specifies an object that corresponds to the jest configuration that should be used in your tests. Stencil provides a default configuration, which you likely won't need to edit, however it can be extended with the same configuration options as Jest. See the [Configuring Jest Guide](https://jestjs.io/docs/en/configuration.html) for configuration details.
+
+:::note
+Keep in mind that the usual way of configuring Jest (`package.json` and `jest.config.js`) is not used with the `stencil testing` command. Jest can still be used, but configuring the presets, transpilation and setting up the correct commands must be done by the project.
+:::
+
+Some additional Stencil specific options may be set here as well for configuring the e2e tests:
+
+```tsx reference title=""
+https://github.com/ionic-team/stencil/blob/278c336166e27d329fffb686f2ecb3185d156696/src/declarations/stencil-public-compiler.ts#L1817-L1963
```
rwaskiewicz, line 19: I don't quite expect us to have an answer to this right now, but want to ask so the team as a whole is thinking about this - how do we maintain this long term? Specifically, if I'm a developer that:
- updates any code highlighted here (so the content has changed and the SHA + line numbers have to change)
- moves this code to another file (so the sha, file path, and perhaps line numbers have to change)

How do I know that this needs to be updated? 

I think this is definitely better than what we have today (a static list that I couldn't tell you when it was last updated without looking at a `git blame`), but do want to start thinking through the next steps here.

docs/testing/stencil-testrunner/02-config.md
```diff
@@ -0,0 +1,20 @@
...
+
+# Testing Config
+
+The `testing` config setting in `stencil.config.ts` specifies an object that corresponds to the jest configuration that should be used in your tests. Stencil provides a default configuration, which you likely won't need to edit, however it can be extended with the same configuration options as Jest. See the [Configuring Jest Guide](https://jestjs.io/docs/en/configuration.html) for configuration details.
+
+:::note
+Keep in mind that the usual way of configuring Jest (`package.json` and `jest.config.js`) is not used with the `stencil testing` command. Jest can still be used, but configuring the presets, transpilation and setting up the correct commands must be done by the project.
+:::
+
+Some additional Stencil specific options may be set here as well for configuring the e2e tests:
```
rwaskiewicz, line 16: Non-blocking - I know this is a copy over from the previous file, but can we end this sentence in a period?

```suggestion
Some additional Stencil specific options may be set here as well for configuring the e2e tests.
```

Right now (to me) it reads if the code is a continuation of just e2e tests, when that's not the case.

docs/testing/stencil-testrunner/_category_.json
```diff
@@ -0,0 +1,5 @@
+{
```
rwaskiewicz, line 1: Non-blocking nit - I think the JSON formatting is a little off here. I'm not opposed to a separate PR to update the Prettier config to check for these things in the future, we simply never checked them before because they so rarely change

docs/testing/stencil-testrunner/_category_.json
```diff
@@ -0,0 +1,5 @@
...
+    "label": "Stencil Testrunner",
```
rwaskiewicz, line 2: Should 'Testrunner' be two words for consistency with the rest of the doc?
```suggestion
    "label": "Stencil Test Runner",
```

package-lock.json
```diff
@@ -6733,6 +6734,14 @@
         "sass": "^1.30.0"
       }
     },
+    "node_modules/docusaurus-theme-github-codeblock": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/docusaurus-theme-github-codeblock/-/docusaurus-theme-github-codeblock-2.0.2.tgz",
+      "integrity": "sha512-H2WoQPWOLjGZO6KS58Gsd+eUVjTFJemkReiSSu9chqokyLc/3Ih3+zPRYfuEZ/HsDvSMIarf7CNcp+Vt+/G+ig==",
+      "dependencies": {
+        "@docusaurus/types": "^3.0.0"
```
rwaskiewicz, line 6742: 👍 I was wondering what this would depend on - love the 1 prod dep

package.json
```diff
@@ -23,6 +23,7 @@
     "@docusaurus/remark-plugin-npm2yarn": "^3.1.0",
     "@ionic-docs/preset-classic": "^1.1.0",
     "clsx": "^2.0.0",
+    "docusaurus-theme-github-codeblock": "^2.0.2",
```
rwaskiewicz, line 26: What happens if we build the site without an internet connection? e.g. What's the failure mode look like if I'm doing a local build and don't have wifi access?

docs/testing/config.md
```diff
@@ -1,160 +0,0 @@
----
```
rwaskiewicz, line 1: For each of the pages we moved/deleted, we'll need to set up redirects in the repo's `vercel.json` file. LMK if you need any help with that!

rwaskiewicz (changes requested):
On additional ask - Can you please also update the acceptance criteria for this ticket?

docs/testing/01-overview.md
```diff
@@ -25,136 +25,14 @@ choose when to use each type of testing:
...
-:::
+- [__Stencil Test Runner__](./stencil-testrunner/01-overview.md): a built-in test runner based on Jest for unit and end-to-end testing with Puppeteer to run within an actual browser in order to provide more realistic results.
 
-The configuration above makes use of special VS Code variables, such as `${workspaceFolder}`.
-These variables are substituted with actual values upon starting the tests.
-For more information regarding the values these variables resolve to, please see VS Code's [Variables Reference documentation](https://code.visualstudio.com/docs/editor/variables-reference).
+:::info
 
-## Other Resources
+We are actively working supporting a more wider range of testing tools, including Playwright and WebdriverIO. Stay tuned for updates!
```
rwaskiewicz, line 36: Non-blocking grammar nit
```suggestion
We are actively working supporting a wider range of testing tools, including Playwright and WebdriverIO. Stay tuned for updates!
```

docs/testing/stencil-testrunner/02-config.md
christian-bromann: It will be hard to put an automatic mechanism in place for updating sha's without manually reviewing the code examples. At least this way it becomes an intentional process. I can see how having thousands of such examples will become out of date quickly and updating them manually will become tedious. However IMHO there is generally no need to update the examples until there is a reason to do so.

docs/testing/stencil-testrunner/_category_.json
christian-bromann: fixed in https://github.com/ionic-team/stencil-site/pull/1344/commits/141f30873c35b5c9f3f2aa59a6c0521ae0908dc2

docs/testing/stencil-testrunner/_category_.json
christian-bromann: fixed in https://github.com/ionic-team/stencil-site/pull/1344/commits/56678eaaacc5732782a12a9269d4c0500ed6bde8

docs/testing/config.md
christian-bromann: No page urls were moved, all still point to the same content / all content remains on the same urls.

Do we want to update the urls and set up redirects instead?

rwaskiewicz: Oh, you're right - that's my fault. I looked at the overview page of the Stencil test runner and assumed that the other pages had changed as well. In that case, I'm OK with leaving them as is - sorry about that!

docs/testing/stencil-testrunner/02-config.md
rwaskiewicz: Agreed - I've spent a bit of time thinking about this myself (before I knew of this project) 😆 😉 

I can see a world where we implement some kind of programmatic check against the current SHA and the primary branch (`main`) and see if the content is different, but I think that's outside the scope of this PR. Just want to see if/what others thought and to plant that idea seed.

docs/testing/config.md
christian-bromann: np!

rwaskiewicz (approved):

christian-bromann:
Should we apply this re-org only to the latest version or to all v4.x and v3.x?

rwaskiewicz:
@christian-bromann I was thinking about that myself. I did a diff of the `testing` subdir in all v4 directories and v3. There are very minute differences (two total) that I think we could simply `cp` the contents of `docs/testing` to each versioned directory and be fine.

Longer term, I'm considering what it'll look like to "squash" all these minor versioned docs into a rolling v4 release. IDK if that'll be my JEDI day project or not next week, but the level of friction to otherwise backport these things is getting higher than I'd like

christian-bromann:
Added commit to deprecated screenshot diffing capabilities:

Screenshot 2024-02-06 at 4 31 06 PM

christian-bromann:
Raised https://github.com/ionic-team/stencil-site/pull/1348 with a massive diff which contains all changes made here to all other versions.

rwaskiewicz:
I'd personally vote we hold off on [deprecating the screenshot functionality](https://github.com/ionic-team/stencil-site/pull/1344/commits/98e31176a9f0b2c674995b6201d7de3e30e90c3c) for right now. I'd prefer we have our docs for WebDriverIO & Playwright available, as well as an initial pass at the Playwright Adapter available so that we can provide folks with a little more guidance.  I don't want this to feel like a "rug pull" for folks that feel they "need" to go find a replacement sooner than later.

rwaskiewicz:
Adding the entire team to this review - I think this is a big enough change that everyone should have a look at what's being proposed here

christian-bromann:
> I'd personally vote we hold off on [deprecating the screenshot functionality](https://github.com/ionic-team/stencil-site/pull/1344/commits/98e31176a9f0b2c674995b6201d7de3e30e90c3c) for right now.

Makes sense. Reverted the change. I am currently working on adding visual testing docs for the new WebdriverIO docs so we have alternatives in the pipeline.

docs/testing/01-overview.md
```diff
@@ -25,136 +25,14 @@ choose when to use each type of testing:
...
-  ]
-}
-```
+Stencil currently supports the following tools for testing components:
 
-:::tip
-Windows users: The `program` value should be set to `"${workspaceFolder}/node_modules/bin/stencil"`.
-If that value does not work, please try`"${workspaceFolder}/node_modules/@stencil/core/bin/stencil"`.
-:::
+- [__Stencil Test Runner__](./stencil-testrunner/01-overview.md): a built-in test runner based on Jest for unit and end-to-end testing with Puppeteer to run within an actual browser in order to provide more realistic results.
```
alicewriteswrongs, line 32: This link doesn't look obviously like a link on the site:

Screenshot 2024-02-07 at 1 27 45 PM

without the bold its "linky-ness" is visually obvious:

Screenshot 2024-02-07 at 1 28 22 PM

Might be a styling issue, but I think unless we're empowered to change the theme we might want to do something like this:

```suggestion
- [Stencil Test Runner](./stencil-testrunner/01-overview.md): a built-in test runner based on Jest for unit and end-to-end testing with Puppeteer to run within an actual browser in order to provide more realistic results.
```

docs/testing/stencil-testrunner/01-overview.md
```diff
@@ -0,0 +1,127 @@
...
+sidebar_label: Overview
+---
+
+# Overview
+
+Stencil has a built-in test runner that uses [Jest](https://jestjs.io/) and [Puppeteer](https://pptr.dev/) as its testing libraries, and allows developers to install both libraries using their preferred package manager.
+
+If you created a project using `npm init stencil`, these libraries were installed for you. Depending on when your project was created, you may or may not have the latest supported version installed.
+
+To view current version support for both Jest and Puppeteer, please see the  [Stencil support policy for testing libraries](../../reference/support-policy.md#testing-libraries).
```
alicewriteswrongs, line 12: Wonder if we might want to add an `info` admonition in here calling out that you need to install `jest` and `puppeteer` if you don't have them - I imagine almost all new Stencil projects are going to be started with `npm init stencil` so users won't run into this, but still could be worth a call-out I think

docs/testing/config.md
alicewriteswrongs: this might be a pain - but if `docs/testing/stencil-testrunner/02-config/md` is the same file that used to be at `docs/testing/config.md` would it be possible to keep our git history around by moving the file using `git mv`?

docs/testing/stencil-testrunner/02-config.md
alicewriteswrongs: this is amazing!

I agree that pointing to a SHA and pulling lines out of a file is much better than what we do now, and I'd be fine with doing this and maybe putting a ticket into the backlog to consider whether we want to add more tooling around this in the future.

If we had a _lot_ of these in the site we could have something which did something like

1. crawl through the `.md` files to find all such references
2. figure out which version the given `.md` is for (i.e. if in `docs` it's `HEAD`, if it's `versioned_docs/version-v4.11` then it's `4.11.x`, etc)
3. grab the git tag in the Stencil repo for that version and do some git wizardry to determine whether the referenced SHA is still the latest commit which changed the referenced lines up to the point of the version that the doc page is specific to
5. if not, fail a CI check on the PR, so that we know we have to update the lines (I think this could be a manual thing, since the starting line and offset could easily change)

That wouldn't proactively push updates to the stencil site when there's a change in the Stencil repo, but I think that 1) eventual consistency is totally fine here and 2) the stencil site does get rebuilt frequently because of edits and dependency updates and whatnot so I don't think it could ever get too far out of date.

However! It's probably a YAGNI / overengineering thing to even really think about tooling like this, especially when this solution + no tooling at all is still a big upgrade over our current copy-paste based solution

package.json
alicewriteswrongs: Is there any way it could store both the information needed to re-fetch the lines from github and a fallback right in the markdown? That would make it possible to build offline without too much trouble I think. I don't know too much about how docusaurus plugins are structured to know if there would be flexibility to do something like this.

alicewriteswrongs (changes requested):
a few questions and comments

docs/testing/01-overview.md
christian-bromann: Yeah, this looks way better. Thanks!

package.json
christian-bromann: I haven't been able to look into this yet but I hope Docusaurus allows to pull the code as part of the server side generation process.

docs/testing/stencil-testrunner/02-config.md
christian-bromann: Sounds like a good Jedi day project 😉

docs/testing/01-overview.md
```diff
@@ -25,136 +25,14 @@ choose when to use each type of testing:
...
-:::
+- [Stencil Test Runner](./stencil-testrunner/01-overview.md): a built-in test runner based on Jest for unit and end-to-end testing with Puppeteer to run within an actual browser in order to provide more realistic results.
 
-The configuration above makes use of special VS Code variables, such as `${workspaceFolder}`.
-These variables are substituted with actual values upon starting the tests.
-For more information regarding the values these variables resolve to, please see VS Code's [Variables Reference documentation](https://code.visualstudio.com/docs/editor/variables-reference).
+:::info
 
-## Other Resources
+We are actively working supporting a wider range of testing tools, including Playwright and WebdriverIO. Stay tuned for updates!
```
tanner-reits, line 36: Minor grammatical fix

```suggestion
We are actively working to support a wider range of testing tools, including Playwright and WebdriverIO. Stay tuned for updates!
```

docs/testing/stencil-testrunner/01-overview.md
tanner-reits, line 12: It might be nice to pull the testing support policy closer to these docs (since we'll be adding most matrices as new adapters come into play), but not needed now

docs/testing/stencil-testrunner/01-overview.md
```diff
@@ -0,0 +1,127 @@
...
+    "scripts": {
+      "test": "stencil test --spec",
+      "test.watch": "stencil test --spec --watchAll",
+      "test.end-to-end": "stencil test --e2e"
+    }
+}
+```
+
+Each command above begins with `stencil test`, which tells Stencil to run tests. Note that each `stencil test` command 
+in example above is followed one or more of the optional flags. Looking at each script, one at a time:
```
tanner-reits, line 36:
```suggestion
in the example above is followed by one or more of the optional flags. Looking at each script, one at a time:
```

docs/testing/stencil-testrunner/03-unit-testing.md
```diff
@@ -14,7 +14,7 @@ To run unit tests, run `stencil test --spec`. Files ending in `.spec.ts` will be
 
 ## newSpecPage()
 
-In order to unit test a component as rendered HTML, tests can use `newSpecPage()` imported from `@stencil/core/testing`. This testing utility method is similar to `newE2EPage()`, however, `newSpecPage()` is much faster since it does not require a full [Puppeteer](https://pptr.dev/) instance to be running. Please see the [newE2EPage()](./e2e-testing.md) docs on more information about complete End-to-end testing with Puppeteer.
+In order to unit test a component as rendered HTML, tests can use `newSpecPage()` imported from `@stencil/core/testing`. This testing utility method is similar to `newE2EPage()`, however, `newSpecPage()` is much faster since it does not require a full [Puppeteer](https://pptr.dev/) instance to be running. Please see the [newE2EPage()](./05-e2e-testing.md) docs on more information about complete End-to-end testing with Puppeteer.
```
tanner-reits, line 17:
```suggestion
In order to unit test a component as rendered HTML, tests can use `newSpecPage()` imported from `@stencil/core/testing`. This testing utility method is similar to `newE2EPage()`, however, `newSpecPage()` is much faster since it does not require a full [Puppeteer](https://pptr.dev/) instance to be running. Please see the [newE2EPage()](./05-e2e-testing.md) docs for more information about complete End-to-end testing with Puppeteer.
```

tanner-reits (approved):
Couple small grammatical fixes, but nothing blocking. Nice job! That Github code block viewer is snazzy 👏

alicewriteswrongs (approved):

PR #1352, opened 2024-02-13
docs(api): add type and example documentation to Stencil primitives

### LINKED ISSUE #1061: forceUpdate in documentation
With v3, it seems forceUpdate has been mostly removed from the documentation.

I did find it on https://stenciljs.com/docs/api#other but the documentation is sparce.

I'm not sure the best update to the documentation, but two points should be made.
* In order to use forceUpdate in a component, you need to import it from @stencil/core.
* In order to call forceUpdate on a component, you need to pass it a reference to a node or element.

For a class component, I used something like the following.

    import { Component, forceUpdate } from '@stencil/core'
    
    // inside a class component function
    forceUpdate(this);

I know that forceUpdate isn't the best solution for production, but it is useful in development and testing.

christian-bromann:
Adding additional docs to our API section.

fixes #1061


docs/components/api.md
```diff
@@ -81,20 +81,104 @@ window.addEventListener('appload', (event) => {
...
+  // inside a class component function
+  forceUpdate(this);
+  ```
+
 - **getAssetPath()**: Gets the path to local assets. Refer to the [Assets](../guides/assets.md#getassetpath) page for usage info.
 
+  __Type:__ `(path: string) => string`<br />
+  __Example:__
+  ```ts
+  import { Component, Prop, getAssetPath } from '@stencil/core'
```
rwaskiewicz, line 136: Same comments as above regarding `h` and the code block type:
```suggestion
  ```tsx
  import { Component, Prop, getAssetPath, h } from '@stencil/core'
```

docs/components/api.md
```diff
@@ -81,20 +81,104 @@ window.addEventListener('appload', (event) => {
...
+      return (
+        <>
+          <div>...</div>
+          <div>...</div>
+          <div>...</div>
+        </>
+      );
+    }
+  }
+  ```
```
rwaskiewicz, line 109: It might be worth calling out here that [jsxFragmentFactory](https://www.typescriptlang.org/tsconfig#jsxFragmentFactory) needs to be set to 'Fragment' here for this to compile. Users will need to add the following to their `tsconfig.json` file:
```diff
diff --git a/tsconfig.json b/tsconfig.json
index 09367ce..c0d45e9 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,7 +14,8 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "jsx": "react",
-    "jsxFactory": "h"
+    "jsxFactory": "h",
+    "jsxFragmentFactory": "Fragment"
   },
   "include": [
     "src"
```

docs/components/api.md
```diff
@@ -81,20 +81,104 @@ window.addEventListener('appload', (event) => {
...
+The following primitives can be imported from the `@stencil/core` package and used within the lifecycle of a component:
+
+- [**Host**](./host-element.md): `<Host>`, is a functional component that can be used at the root of the render function to set attributes and event listeners to the host element itself. Refer to the [Host Element](./host-element.md) page for usage info.
+
+- **Fragment**: `<Fragment>`, often used via `<>...</>` syntax, lets you group elements without a wrapper node.
+
+  __Type:__ `FunctionalComponent`<br />
+  __Example:__
+  ```ts
+  import { Component, Fragment } from '@stencil/core'
```
rwaskiewicz, line 93: Should we add `h` here for the `div` elements?
```suggestion
  import { Component, Fragment, h } from '@stencil/core'
```

rwaskiewicz (changes requested):

versioned_docs/version-v4.12/components/api.md
```diff
@@ -81,7 +81,9 @@ window.addEventListener('appload', (event) => {
 
 ## Other
 
-- [**Host**](./host-element.md): Host is a functional component that can be used at the root of the render function to set attributes and event listeners to the host element itself.
+The following primitives can be imported from the `@stencil/core` package and used within the lifecycle of a component:
```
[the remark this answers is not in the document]
christian-bromann: Good catch 👌

christian-bromann:
@rwaskiewicz thanks for the feedback, let me know if I addressed your comments correctly.

versioned_docs/version-v4.12/components/api.md
```diff
@@ -81,20 +81,104 @@ window.addEventListener('appload', (event) => {
...
+
+- [**Host**](./host-element.md): `<Host>`, is a functional component that can be used at the root of the render function to set attributes and event listeners to the host element itself. Refer to the [Host Element](./host-element.md) page for usage info.
+
+- **Fragment**: `<Fragment>`, often used via `<>...</>` syntax, lets you group elements without a wrapper node. Make sure you define a `jsxFragmentFactory` compiler option alongside `jsxFactory` when using this functional component.
```
rwaskiewicz, line 88: Can we clarify which compiler (stencil vs typescript) this is for? 

I tried to add the word "TypeScript" along with the value setting to "Fragment" and my edits didn't sound good in my head. What do you think about:

```suggestion
- **Fragment**: `<Fragment>`, often used via `<>...</>` syntax, lets you group elements without a wrapper node.

To use this feature, ensure that the following TypeScript compiler options are set:
- [`jsxFragmentFactory` is set](https://www.typescriptlang.org/tsconfig#jsxFragmentFactory) to "Fragment"
- [`jsxFactory` is set](https://www.typescriptlang.org/tsconfig#jsxFactory) to "h"
```

rwaskiewicz (approved):

versioned_docs/version-v4.12/components/api.md
```diff
@@ -81,20 +81,108 @@ window.addEventListener('appload', (event) => {
...
-- [**Host**](./host-element.md): Host is a functional component that can be used at the root of the render function to set attributes and event listeners to the host element itself.
+The following primitives can be imported from the `@stencil/core` package and used within the lifecycle of a component:
+
+- [**Host**](./host-element.md): `<Host>`, is a functional component that can be used at the root of the render function to set attributes and event listeners to the host element itself. Refer to the [Host Element](./host-element.md) page for usage info.
+
+- **Fragment**: `<Fragment>`, often used via `<>...</>` syntax, lets you group elements without a wrapper node.
+
+To use this feature, ensure that the following TypeScript compiler options are set:
+- [`jsxFragmentFactory` is set](https://www.typescriptlang.org/tsconfig#jsxFragmentFactory) to "Fragment"
+- [`jsxFactory` is set](https://www.typescriptlang.org/tsconfig#jsxFactory) to "h"
```
alicewriteswrongs, line 92: I don't see these lines in the other versions, so I think maybe it just needs to be copied around

alicewriteswrongs (approved):
looks good! Just noticed that one thing needs to be copied around

versioned_docs/version-v4.12/components/api.md
christian-bromann: Good catch 👌 I populated the change to other versions

PR #1354, opened 2024-02-15
docs(props): clarify runtime types of union props

rwaskiewicz:
extend the section on union types to clarify the runtime type of a union type at runtime. previously, it was difficult to understand/parse the existing docs to determine how a `@Prop`'s runtime type was determined when used in a running library or application. this commit intends to make this a bit clearer, especially in the case of `string|boolean` types being used in HTML and TSX


docs/components/properties.md
```diff
@@ -567,6 +567,61 @@ and TSX:
 <todo-list-item isComplete={false}></todo-list-item>
 ```
 
+When using union types, the type of a component's `@Prop()` value can be ambiguous at runtime.
+In the example above, when is `@Prop() isComplete` a `string`, and when is it a `boolean`?
```
christian-bromann, line 571: I could a little bit stuck in reading, how about this wording?
```suggestion
In the provided example, under what circumstances does `@Prop() isComplete` function as a `string`, and when does it serve as a `boolean`?
```

christian-bromann (approved):
One comment, otherwise LGTM 👍 do we want to back port this to the v3 docs as well?

rwaskiewicz:
@christian-bromann 

> do we want to back port this to the v3 docs as well?

Yeah, I'll do that once this gets another sign off before I merge it

docs/components/properties.md
```diff
@@ -567,6 +567,61 @@ and TSX:
...
+```html
+<!-- Since this is HTML, the value of `isComplete` in `ToDoListItem` will be a string -->
+
+<!-- Set isComplete to "true". -->
+<todo-list-item is-complete="true"></todo-list-item>
+<!-- Set isComplete to 'false' -->
+<todo-list-item is-complete="false"></todo-list-item>
+```
+
+Likewise, when the attribute on a component is set, the runtime value of a `@Prop()` is always a string.
```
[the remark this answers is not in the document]
alicewriteswrongs: Maybe we should also just say that if `typeof val` is _not_ `"string"` then it's coerced to a string, we could link to this resource to clarify if people are (understandably) not familiar with every detail of how this happens: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#string_coercion

Might help avoid the dreaded `[object Object]`

docs/components/properties.md
```diff
@@ -567,6 +567,61 @@ and TSX:
...
+  // both of these `setAttribute` calls set the property `isComplete` to "true" (string)
+  document.querySelector('todo-list-item').setAttribute('is-complete', true);
+  document.querySelector('todo-list-item').setAttribute('is-complete', "true");
+  // both of these `setAttribute` calls set the property `isComplete` to "false" (string)
+  document.querySelector('todo-list-item').setAttribute('is-complete', false);
+  document.querySelector('todo-list-item').setAttribute('is-complete', "false");
+</script>
+```
+
+However, if the property of a custom element is directly changed, the `typeof` of the `@Prop()` provided is used.
```
alicewriteswrongs, line 596:
```suggestion
However, if the property of a custom element is directly changed by accessing the setter on the element, the supplied value is used without performing any type conversion. In other words, the `typeof` the value will remain the same.
```

Possibly change, just to break down a bit what is going on (getting the value in via the setter / getter or defined property as opposed to the attribute)

docs/components/properties.md
```diff
@@ -567,6 +567,61 @@ and TSX:
 <todo-list-item isComplete={false}></todo-list-item>
 ```
 
+When using union types, the type of a component's `@Prop()` value can be ambiguous at runtime.
+In the provided example, under what circumstances does `@Prop() isComplete` function as a `string`, and when does it serve as a `boolean`?
+
+When using a component in HTML, the runtime value of a `@Prop()` is always a string.
+This is a result of setting the HTML attribute for the custom element.
```
alicewriteswrongs, line 574: Should we possibly mention that this is in-line with how attributes work on html elements in general? I don't know, might be an opportunity to do a little web fundamentals education for the reader which is helpful for understanding the 'why' here

docs/components/properties.md
```diff
@@ -567,6 +567,61 @@ and TSX:
...
+  // Set the property `isComplete` to "true" (string)
+  document.querySelector('todo-list-item').isComplete = "true";
+  // Set the property `isComplete` to `false` (boolean)
+  document.querySelector('todo-list-item').isComplete = false;
+  // Set the property `isComplete` to "false" (string)
+  document.querySelector('todo-list-item').isComplete = "false";
+</script>
+```
+
+When using a component in TSX, the `typeof` of the `@Prop()` provided is used.
```
alicewriteswrongs, line 610: Maybe we should mention this is similar behavior to using the property?

alicewriteswrongs (changes requested):
Looks good overall! Good to clarify this stuff. Just had a few suggestions and thoughts

docs/components/properties.md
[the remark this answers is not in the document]
rwaskiewicz: Sounds good to me! I applied most of these changes in 5ab412d72 - the only thing I held off on was adding the "from JavaScript" bit - I'm a little concerned that folks will confuse calling `setAttribute` in a `<script>` tag in an HTML with calling `setAttribute` in a JavaScript _file_. While this behavior applies to both, I didn't want to imply calling `setAttribute` in a `script` tag in an HTML file would be different in some way

docs/components/properties.md
rwaskiewicz: I think this might be a little too in the weeds for folks - what do you think about something like 6648bdc5a? I tried to simplify what I had previously here to try say "What you give is what you get" 😆

docs/components/properties.md
rwaskiewicz: I took a stab at this in 6648bdc5a - LMK what you think!

docs/components/properties.md
rwaskiewicz: I go back and forth here - on one hand, I think it'd be beneficial to give folks some background knowledge that are coming to Stencil/WC in general. On the other, I don't think it's our responsibility to teach HTML to folks. Back on the original hand, there's some trickiness with Stencil/TS and marrying some of those types up with HTML types.

I made some copy edits in aa91e1e39 - LMK what you think. I held off on mentioning the spec/talking about "standards" too much here, although I think there's some room for improvement.

Maybe this aligns well with some of the thoughts/things @tanner-reits was talking about during our last JEDI days for reorganizing the docs a bit - specifically I wonder if we should revisit our onboarding experience to consider some WC fundamentals needed to use Stencil effectively 🤔 

Since don't align with the spec in some cases, and I can see a world where saying "Here's the spec, here's how to work effectively using it, and here are the special things Stencil does" having some value for users. Thoughts? (Out of scope for this PR obviously 🙃)

docs/components/properties.md
alicewriteswrongs: sure, I think that's fine, just thinking we want to clarify we're talking about runtime types which means possible type conversion

docs/components/properties.md
alicewriteswrongs: Sure, that edit looks good to me - my thought on mentioning this is part of attributes in general is just providing a bit of the 'why' for this dichotomy in behavior b/tw attrs specified in HTML vs using `setAttribute`, I've always found that if you have a 'why' explanation it's easier to remember something like this as opposed to if it's just presented as sort of an arbitrary 'this is how it is.' But I think the edit is good as is.

On the larger rewrite I think that's a good idea, in general it would be good I think for developers to have a good intuition about where Stencil ends and custom elements begin, I don't really know how much that would be understood by the average Stencil user at present but if our docs could be structured to facilitate that understanding I think that would be a step forward

PR #1359, opened 2024-02-22
docs(publishing): update general docs on publishing

christian-bromann:
Currently our documentation around publishing docs in not very well structured and content is scattered among different pages. This caused confusions in the ecosystem. This patch attempts to clean up our docs by restructuring the content and put an emphasis on the distribution use cases: lazy loading and standalone Stencil modules. To summarize the changeset:

- I removed content on the topic of publishing components within the output targets docs and moved parts of it into the general publishing docs
- I updated the structure of our publishing docs with focus on the use cases
  - moved the recommendation for `package.json` exports based on the use case
  - added a section on consideration to explain when a certain use case might be favorable over the other


docs/guides/publishing.md
```diff
@@ -5,52 +5,134 @@ description: Publishing A Component Library
...
+}
+```
+
+:::note
+If you set the `dir` property on the output target config, replace `dist/components` in the above snippet with the path set in the config.
+:::
+
+## Publishing to NPM
+
+The first step and highly recommended step is to [publish the component library to NPM](https://docs.npmjs.com/getting-started/publishing-npm-packages). [NPM](https://www.npmjs.com/) is an online software registry for sharing libraries, tools, utilities, packages, etc. Once the library is published to NPM, other projects are able to add your component library as a dependency and use the components within their own projects.
```
rwaskiewicz, line 138: I wonder if we should reword this a bit:
```suggestion
[NPM](https://www.npmjs.com/) is an online software registry for sharing libraries, tools, utilities, packages, etc. To make your Stencil project widely available to be consumed, it's recommended to [publish the component library to NPM](https://docs.npmjs.com/getting-started/publishing-npm-packages). Once the library is published to NPM, other projects are able to add your component library as a dependency and use the components within their own projects.
```

Thoughts?

rwaskiewicz (changes requested):

docs/guides/publishing.md
```diff
@@ -5,52 +5,134 @@ description: Publishing A Component Library
...
-    "css/",
-    "loader/"
-  ]
+  "exports": "./dist/esm/my-design-system.js",
+  "main": "./dist/cjs/my-design-system.js",
+  "unpkg": "dist/my-design-system/my-design-system.esm.js",
+}
+```
+
+Read more about various options when it comes to distributing your components lazily in the [`dist`](../output-targets/distribution.md) output target section.
```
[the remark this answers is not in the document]
christian-bromann: Yeah, this is unclear. I will change it to:

```suggestion
Read more about various options when it comes to configuring your project's components for lazy loading in the [`dist`](../output-targets/distribution.md) output target section.
```

docs/guides/publishing.md
christian-bromann: Looks good!

christian-bromann:
Thanks for the feedback. I applied the updates accordingly.

rwaskiewicz (changes requested):
Thanks! I had a few more asks after taking some time between reads. LMK if you have any questions

docs/guides/publishing.md
```diff
@@ -5,52 +5,134 @@ description: Publishing A Component Library
...
+
+```ts
+import { MyComponent, defineCustomElementMyComponent } from 'my-design-system'
+
+// register to CustomElementRegistry
+defineCustomElementMyComponent()
+
+// or extend custom element via
+class MyCustomComponent extends MyComponent {
+  // ...
```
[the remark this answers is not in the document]
christian-bromann: Yes, for example, let's say we publish a `create-stencil` component, you can change functionality by extending the class and overwriting component methods, e.g.:

```html
<script type="module">
  import { MyComponent } from 'https://unpkg.com/foobar-design-system@0.0.4/dist/components/index.js'
  class ChangedComponent extends MyComponent {
    getText() {
      return 'a new component!';
    }
  }

  customElements.define('my-component', ChangedComponent);
  customElements.define('my-component-orig', MyComponent);
</script>
<my-component first="foo" last="bar"></my-component>
<my-component-orig first="foo" last="bar"></my-component-orig>
```

This will result in:

Screenshot 2024-02-28 at 10 16 25 AM

docs/output-targets/custom-elements.md
```diff
@@ -128,32 +128,6 @@ _default: `false`_
 
 Setting this flag to `true` will cause file minification to follow what is specified in the [Stencil config](../config/01-overview.md#minifyjs). _However_, if [`externalRuntime`](#externalruntime) is enabled, it will override this option and always result in minification being disabled.
 
-## Consuming Custom Elements
-
-By default, the custom elements files will be written to `dist/components/`. This directory can be configured using the output target's [`dir`](#dir) config.
```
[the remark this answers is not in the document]
christian-bromann: I added some sections back in `Publishing > Use Cases > Standalone`

docs/guides/publishing.md
```diff
@@ -5,52 +5,155 @@ description: Publishing A Component Library
...
 
-The first step and highly recommended step is to
-[publish the component library to NPM](https://docs.npmjs.com/getting-started/publishing-npm-packages). NPM is an online software registry for sharing libraries, tools, utilities, packages, etc. Once the library is published to NPM, other projects are able to add your component library as a dependency and use the components within their own projects.
+### Lazy Loading
 
+If you prefer to have your components automatically loaded when used in your application, we recommend to enable the [`dist`](../output-targets/dist.md) output target. The bundle gives you a small entry file that registers all your components and only loads the full component logic when it gets rendered in your application. It doesn't matter if the actual application is written in HTML or created with vanilla JavaScript, jQuery, React, etc.
```
alicewriteswrongs, line 16:
```suggestion
If you prefer to have your components automatically loaded when used in your application, we recommend enabling the [`dist`](../output-targets/dist.md) output target. The bundle gives you a small entry file that registers all your components and defers loading the full component logic until it is rendered in your application. It doesn't matter if the actual application is written in HTML or created with vanilla JavaScript, jQuery, React, etc.
```

just some wording suggestions here

docs/guides/publishing.md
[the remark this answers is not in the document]
alicewriteswrongs: wow I didn't know that worked! I guess I hadn't tried it

docs/guides/publishing.md
```diff
@@ -5,52 +5,155 @@ description: Publishing A Component Library
...
+  "exports": {
+    ".": {
+      "import": "./dist/components/index.js",
+      "types": "./dist/components/index.d.ts"
+    },
+    "./my-component": {
+      "import": "./dist/components/my-component.js",
+      "types": "./dist/components/my-component.d.ts"
+    }
+  },
```
alicewriteswrongs, line 82: not something we should do now, but it would be nice for supporting this usage of `exports` with DCE if we could auto-generate a JSON blob for all the components, so that users didn't have to automatically keep it in sync - imagine the nightmare of doing that if you had 200 components or something like that

docs/guides/publishing.md
```diff
@@ -5,52 +5,155 @@ description: Publishing A Component Library
...
-The `collection` properties are used to allow lazy loading in other Stencil applications.
+If you define exports targets for all your components as shown above and by using [`customElementsExportBehavior: 'auto-define-custom-elements'`](../output-targets/custom-elements.md#customelementsexportbehavior) as output target option, you can skip the `defineCustomElement` call and directly import the component where you need it:
+
+```ts
+import 'my-design-system/my-component'
+```
+
+:::note
+If you are distributing both the `dist` and `dist-custom-elements`, then it's best to pick one of them as the main entry depending on which use case is more prominent.
+:::
```
alicewriteswrongs, line 104: By this you mean the file path in the `main` field in `package.json`? There's a `main` field shown above for `dist` but it doesn't show one for DCE, might be helpful to have that more explanation of what setting the `main` field would look like for DCE

(possible I'm also misunderstanding something here haha)

alicewriteswrongs (changes requested):
just had a few questions and suggestions, looks like a step forward overall!

docs/guides/publishing.md
christian-bromann: What does DCE stands for?

christian-bromann: Yeah, sounds like a great idea!

docs/guides/publishing.md
christian-bromann: I left this out intentionally as I don't know if it is a good idea to define a `main` entry point DCE. `main` is usually only used for CJS environments. We don't export DCE as CJS module. I am sure this can be configured but don't think anyone would want to import CJS to then re-transpile back to ESM which is what is supported in the browser.

alicewriteswrongs: Ok that sounds fine, I guess if the user is going to distribute _both_ then we can rely on them a bit to figure out how to do that correctly

alicewriteswrongs (approved):

docs/guides/publishing.md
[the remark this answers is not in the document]
tanner-reits: Yeah extending the compiled class output will work, but obviously wouldn't be able to add any additional reactivity (states, props, watchers, etc.) at this time.

tanner-reits (approved):

tanner-reits (approved):

PR #1362, opened 2024-02-26
docs(copy-tasks): rewrite documentation for copy tasks

### LINKED ISSUE #852: website docs for Copy Tasks for Output Targets are incorrect
**Stencil version:**

```
 @stencil/core@1.17.3
```

**I'm submitting a:**

[ x ] bug report

**Current behavior:**

The [docs](https://stenciljs.com/docs/copy-tasks) for copy tasks says:
```Each output target can have its own copy config, which is an array of objects that defines any files or folders that should be copied over to the output target's build directory.```

But `docs` target does not support `copy`.

**Expected behavior:**

Docs to not say that each output target can have a copy config.
Or docs to support copy task.

**Steps to reproduce:**

Type error if you add `copy` to output target with type `docs` in stencil.config.ts

alicewriteswrongs:
This updates the documentation for the `copy` configuration that can be made on the `www`, `dist-custom-elements`, and `dist` output targets. This was motivated by wanting to clarify that this option is not supported on the documentation-related output targets (e.g. `docs-json`).

A change was recently made in Stencil to document the related interface, which we can take advantage of here:
https://github.com/ionic-team/stencil/pull/5413

fixes #852
STENCIl-1047

### DIFF
--- a/docs/output-targets/copy-tasks.md
+++ b/docs/output-targets/copy-tasks.md
@@ -8,13 +8,26 @@ slug: /copy-tasks
 
 # Copy Tasks for Output Targets
 
-Each output target can have its own `copy` config, which is an array of objects that defines any files or folders that should be copied over to the output target's build directory.
+All of Stencil's non-documentation output targets
+([`dist-custom-elements`](./custom-elements.md), [`dist`](./dist.md), and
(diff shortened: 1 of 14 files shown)


christian-bromann (approved):
👍

tanner-reits (approved):

PR #1370, opened 2024-03-08
Update 01-overview.md

eriklharper:
correcting spelling error on line 40


alicewriteswrongs (changes requested):
thanks for finding this typo!

would you mind also updating the file at `docs/output-targets/01-overview.md`? the docs in `docs/` are the 'next' docs which will be used as the basis for future versions, so we want to make sure changes like this are made in there as well so that they'll be applied to future versions of the docs.

If you're up for it you could also fix this in earlier versions too! It looks like we've had the typo for a while.

eriklharper:
https://github.com/ionic-team/stencil-site/pull/1370

rwaskiewicz (approved):

alicewriteswrongs (approved):

PR #1373, opened 2024-03-13
docs(testing): add caution for piercing shadow DOM

christian-bromann:
We have seen in Stencil that user are trying to use multiple `>>>` selectors within a single query which is afaik not possible. This patch adds docs to emphasize how to work around this limitation.


docs/testing/stencil-testrunner/05-e2e-testing.md
```diff
@@ -85,6 +85,20 @@ Use the "piercing" selector `>>>` to query for an object inside a component's sh
 const el = await page.find('foo-component >>> .close-button');
 ```
 
+:::caution
+Make sure to only use `>>>` once in your query. Puppeteer does not support nested deep selectors. Instead, separate the calls if possible into two queries, e.g.:
```
tanner-reits, line 89: Minor suggestion:
```suggestion
Make sure to only use `>>>` once in your query. Puppeteer does not support nested deep selectors. Instead, separate the calls into two queries, e.g.:
```

tanner-reits (approved):
1 non blocker

rwaskiewicz (approved):
I'm guessing some of this will be replaced with https://github.com/ionic-team/stencil/pull/5481, but good to get the documentation in for older versions of Stencil 👍

christian-bromann:
> I'm guessing some of this will be replaced with [ionic-team/stencil#5481](https://github.com/ionic-team/stencil/pull/5481)

Giving the limitations from the piercing approach using multiple `>>>` in one query I would advise this approach over the one suggested in #5481

PR #1375, opened 2024-03-14
docs(support): add support note for TypeScript 5.4 in Stencil v4.13 and higher

christian-bromann:
Update support policy for TypeScript 5.4 support.

### DIFF
--- a/docs/reference/support-policy.md
+++ b/docs/reference/support-policy.md
@@ -86,6 +86,7 @@ The table below describes recent versions of Stencil and the version of TypeScri
 
 | Stencil Version | TypeScript Version |
 |:---------------:|:------------------:|
+|     v4.13.0     |       v5.4.0       |
 |     v4.10.0     |       v5.3.0       |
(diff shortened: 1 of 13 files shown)


docs/reference/support-policy.md
```diff
@@ -86,6 +86,7 @@ The table below describes recent versions of Stencil and the version of TypeScri
 
 | Stencil Version | TypeScript Version |
 |:---------------:|:------------------:|
+|     v4.13.0     |       v5.4.0       |
```
rwaskiewicz, line 89: I think we only need to change this file - historically we don't backfill the version support doc. Can we reset the other versioned docs to their previous state?

rwaskiewicz (changes requested):

docs/reference/support-policy.md
christian-bromann: I've done it with the last TS update but went back and removed these references again.

rwaskiewicz (approved):

tanner-reits (approved):

PR #1378, opened 2024-03-18
docs(testing): add initial playwright adapter docs

tanner-reits:
Adds initial documentation for the Stencil Playwright adapter.

I tried to keep this documentation lightweight and simple for now. This will definitely evolve over time as we get more users and feedback on areas to clarify/improve. I also tried to point to Playwright's official docs as much as possible since their APIs are really what are getting used at the end of the day.


rwaskiewicz:
Team - please use `@stencil/playwright@0.1.0-dev.1710793791.3353b57` to test these docs out. It's the latest dev build from the head of main. Note, you'll need a project with Stencil v4.13.0+

docs/testing/playwright/02-e2e-testing.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+## Running Tests
+
+By default, the adapter will execute all tests in a project with a `.e2e.ts` file suffix. This can be modified by passing the
+[`testDir`](https://playwright.dev/docs/api/class-testproject#test-project-test-dir) and/or [`testMatch`](https://playwright.dev/docs/api/class-testproject#test-project-test-match)
+configuration options as overrides to `createStencilPlaywrightConfig()`.
+
+To run tests, either run `npx playwright test` from the **root** of the Stencil project, or update the project's `test` script in the `package.json` file to run the
+Playwright command.
```
rwaskiewicz, line 111: What do you think about switching the order of these two sections? I don't want to imply that `npx stencil test` will run these playwright tests, and am thinking it'd make sense to say "run `npx playwright test` right out of the gate:

```suggestion
To run tests, either run `npx playwright test` from the **root** of the Stencil project, or update the project's `test` script in the `package.json` file to run the
Playwright command.

By default, the adapter will execute all tests in a project with a `.e2e.ts` file suffix. This can be modified by passing the
[`testDir`](https://playwright.dev/docs/api/class-testproject#test-project-test-dir) and/or [`testMatch`](https://playwright.dev/docs/api/class-testproject#test-project-test-match)
configuration options as overrides to `createStencilPlaywrightConfig()`.
```

rwaskiewicz (changes requested):
Did a first pass at reviewing. Overall looks good, most of my comments were minor. I'll circle back tomorrow and test some of these examples out.

Also, it looks like the build is failing due to broken links.

docs/testing/playwright/01-overview.md
```diff
@@ -0,0 +1,63 @@
...
+1. Create a Playwright config at the root of your Stencil project:
+
+   ```ts title="playwright.config.ts"
+   import { expect } from '@playwright/test';
+   import { matchers, createStencilPlaywrightConfig } from '@stencil/playwright';
+
+   // Add custom Stencil matchers to Playwright assertions
+   expect.extend(matchers);
+
+   export default createStencilPlaywrightConfig();
```
christian-bromann, line 42: Maybe show case that folks can extend the PW config here:
```suggestion
   export default createStencilPlaywrightConfig({
       // overwrite Playwright config options here
       reporter: 'line',
       // ...
   });```

Comment (christian-bromann) [reply docs/testing/playwright/01-overview.md:42]: Are there any configs that aren't allowed to be overwritten, people should be aware of?

**Review by christian-bromann — APPROVED** [review]
One minor comment (optional)

LGTM 👍

File: docs/testing/playwright/02-e2e-testing.md [hunk 3]
```diff
@@ -0,0 +1,125 @@
...
+
+```html title="my-component.e2e.html"
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf8" />
+
+    <!-- Replace with the path to your entrypoint -->
+    <script src="./build/test-app.esm.js" type="module"></script>
+    <script src="./build/test-app.js" nomodule></script>
```
rwaskiewicz, line 42: Are these paths relative to `www`? I'm having trouble getting a 'hello world' style test setup using this example here, and suspect it's because I'm not loading these correctly.

For a starter created with `npm init stencil@latest component playwright-test`:
```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf8" />

    <!-- Replace with the path to your entrypoint -->
    <script src="./wbuild/playwright-test.esm.js" type="module"></script>
    <script src="./build/playwright-test.js" nomodule></script>
  </head>
  <body>
    <my-component first="Stencil"></my-component>
  </body>
</html>
```
and 
```ts
import {expect} from '@playwright/test'
import {test} from '@stencil/playwright';

test.describe('my-component', () => {
  test('should render the correct name', async ({page}) => {
    await page.goto('my-component.e2e.html');
    expect(true).toBe(true)
  })
})
```

This times out after 30 secounds on the `goto`:
```
  1) src/components/my-component/my-component.e2e.ts:5:7 › my-component › should render the correct name

    Test timeout of 30000ms exceeded.

    Error: page.waitForFunction: Test timeout of 30000ms exceeded.

      4 | test.describe('my-component', () => {
      5 |   test('should render the correct name', async ({page}) => {
    > 6 |     await page.goto('/my-component/my-component.e2e.html');
        |                ^
      7 |     expect(true).toBe(true)
      8 |   })
      9 | })
```

Is there anything I'm missing that's blatantly obvious here? If I run this with `--headed` mode enabled, I get
```
404 File Not Found
Url: /my-component/my-component.e2e.html
File: /private/tmp/playwright-test/www/my-component/my-component.e2e.html
```
where the location of `my-component.e2e.html` on disk is actually `/private/tmp/playwright-test/src/components/my-component/my-component.e2e.html` - should my HTML file be under `www`?

docs/testing/playwright/02-e2e-testing.md
```diff
@@ -0,0 +1,125 @@
...
+or if additional `script` or `style` tags need to be included in the HTML. However, with this pattern, developers are responsible for defining the necessary `script`
+tags pointing to the Stencil entry code (so all web components are correctly loaded and registered).
+
+```html title="my-component.e2e.html"
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf8" />
+
+    <!-- Replace with the path to your entrypoint -->
```
rwaskiewicz, line 40: Do we know if folks will know that an 'entrypoint' is in this context? Is there anything we could do to clarify what 'test-app' should be in their project?

docs/testing/playwright/03-api.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Signature:** `createStencilPlaywrightConfig(overrides?: CreateStencilPlaywrightConfigOptions): Promise<PlaywrightTestConfig>`
+
+Returns a [Playwright test configuration](https://playwright.dev/docs/test-configuration#introduction).
+
+`overrides`, as the name implies, will overwrite the default configuration value(s) if supplied. These values can include any valid Playwright config option as well
+as some options to override specific values in the config options related to the Stencil integration:
+
+- `webServerCommand`: This can be specified to change the command that will run to start the Stencil dev server. Defaults to `npm start -- --no-open`.
+- `webServerTimeout`: This can be specified to change the amount of time Playwright will wait for the dev server to start. Defaults to 60 seconds.
```
rwaskiewicz, line 16: Can we clarify the unit of this configuration options is millis? I think it can get a little confusing since we say it defaults to 60 seconds, then use millis in the example.
```suggestion
- `webServerTimeout`: This can be specified to change the amount of time Playwright will wait for the dev server to start in milliseconds. Defaults to 60000 (60 seconds).
```

rwaskiewicz (changes requested):
Few more asks/questions after playing around with this some more and reading it with a clearer head

docs/testing/playwright/01-overview.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Playwright Overview
+sidebar_label: Overview
+---
+
+Playwright is an automated end-to-end testing framework built to run on all modern browser engines and operating systems. Playwright leverages
+the DevTools protocol to provide reliable tests that run in actual browsers.
```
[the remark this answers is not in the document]
christian-bromann: Yes it does .. in fact this is the reason why they patch browser so they extend or make these devtools protocol e.g. in Safari accessible

docs/testing/playwright/01-overview.md
tanner-reits: Nope, we allow all to be overwritten. That being said, there are some I'd advise against overwriting, so I'll add something around that.

docs/testing/playwright/01-overview.md
```diff
@@ -0,0 +1,63 @@
...
+
+   The `createStencilPlaywrightConfig()` is a utility that will create a default Playwright configuration based on your project's Stencil config. Read
+   more about how to use this utility in the [API docs](./03-api.md#createstencilplaywrightconfig-function).
+
+1. Ensure the Stencil project has a [`www` output target](../../output-targets/www.md). Playwright relies on pre-compiled output running in a dev server
+   to run tests against. When using the `createStencilPlaywrightConfig()` helper, a configuration for the dev server will be automatically created based on
+   the Stencil project's `www` output target config and [dev server config](../../config/dev-server.md). If no `www` output target is specified,
+   tests will not be able to run.
+
+1. It _may_ be necessary to update your project's `tsconfig.json` to add the `ESNext.Disposable` option to the `lib` array:
```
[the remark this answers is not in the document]
tanner-reits: Yeah that's clearer. Might trim that down to just say something along the lines of "..., you may hit an error message saying "Property 'asyncDispose' does not exist on type 'SymbolConstructor'" ..."

docs/testing/playwright/02-e2e-testing.md
tanner-reits: Good call out. Totally meant to try and clarify and guide a bit more here cuz this is probably the trickiest bit in the whole adapter. I'll throw something together

tanner-reits: Added some text to clarify the expected values from the code snippets. Let me know if that is clear enough or if there is more I can add.

docs/testing/playwright/02-e2e-testing.md
```diff
@@ -0,0 +1,133 @@
...
+As far as writing the actual tests goes, the most important thing to be aware of for tests to run correctly is to import the `test` function from
+`@stencil/playwright`, **not** directly from `@playwright/test`:
+
+```ts
+// THIS IS CORRECT
+import { test } from '@stencil/playwright';
+
+// THIS IS NOT!!
+// import { test } from '@playwright/test';
+```
```
alicewriteswrongs, line 21: I know it might be more work, but is there any way we could have a minimal, smallest-possible 'hello world' type test here? I think with any tutorial like this one of the most important things is getting the reader to a state where they have _something_ running as quickly as possible.

But when I'm reading through it's not immediately apparent what I want to do here to go forward. If I go check the ['Writing Tests'](https://playwright.dev/docs/writing-tests) page on the Playwright docs I see this:

```ts
import { test, expect } from '@playwright/test';

test('has title', async ({ page }) => {
  await page.goto('https://playwright.dev/');

  // Expect a title "to contain" a substring.
  await expect(page).toHaveTitle(/Playwright/);
});
```

but I'm not sure how exactly I'd want to adapt that to make it work with our setup, beyond that it looks like I need to `import { test } from '@stencil/playwright'`. But what do I want to pass to `page.goto`?

docs/testing/playwright/01-overview.md
```diff
@@ -0,0 +1,66 @@
...
+   });
+   ```
+
+   The `createStencilPlaywrightConfig()` is a utility that will create a default Playwright configuration based on your project's Stencil config. Read
+   more about how to use this utility in the [API docs](./03-api.md#createstencilplaywrightconfig-function).
+
+1. Ensure the Stencil project has a [`www` output target](../../output-targets/www.md). Playwright relies on pre-compiled output running in a dev server
+   to run tests against. When using the `createStencilPlaywrightConfig()` helper, a configuration for the dev server will be automatically created based on
+   the Stencil project's `www` output target config and [dev server config](../../config/dev-server.md). If no `www` output target is specified,
+   tests will not be able to run.
```
alicewriteswrongs, line 53: do I need any specific `devServer` config? I have this Stencil config:

```ts
import { Config } from '@stencil/core';

  namespace: 'test-playwright',
  outputTargets: [
    {
      type: 'dist',
      esmLoaderPath: '../loader',
    },
    {
      type: 'dist-custom-elements',
    },
    {
      type: 'docs-readme',
    },
    {
      type: 'www',
      serviceWorker: null, // disable service workers
    },
  ],
  testing: {
    browserHeadless: "new",
  },
  devServer: {
  }
};
```

and when I do `npx playwright` I just see a timeout, like this:

Screenshot 2024-03-21 at 10 13 33 AM

Maybe I did something wrong! Probably!

I only have one test file I created which looks like this:

```ts
import { expect } from '@playwright/test';
import { test } from '@stencil/playwright';

test.describe('my-component', () => {
  test('should render the correct name', async ({ page }) => {
    // Assume we have a template setup with the `my-component` component and a `button`
    // await page.goto('/my-component/my-component.e2e.html');
    await page.setContent('<my-component first="Stencil"></my-component>');

    const button = page.locator('button');
    // Assmume clicking the button changes the `@Prop()` values on `my-component`
    button.click();
    await page.waitForChanges();
  });
});
```

it's saved in `src/my-component/my-component.e2e.ts`

docs/testing/playwright/02-e2e-testing.md
alicewriteswrongs: Actually thinking about it more, I think it might be good to have a minimal test on the first page ('Overview') as the last step, what do you think?

alicewriteswrongs (changes requested):
I ran into a few roadblocks, which might be of my own making!

docs/testing/playwright/02-e2e-testing.md
tanner-reits: I want to create a live code example on StackBlitz we can link out to for a working example

alicewriteswrongs: makes sense!

docs/testing/playwright/01-overview.md
tanner-reits: @alicewriteswrongs So this is one bit I don't like, this is caused by the `Symbol.asyncDispose` issue. I bet if you run `npm run build`, you'll see an error. The problem is that Playwright (or our implementation), doesn't report build errors when running the dev server, so nothing fails until the timeout is reached.

I have a task in Jira to figure out a way for us to log those errors and "fail quick" when they happen.

For now, I'll update the step below this one to always recommend adding the `ESNext.Disposable` property in the `tsconfig.json` since this is a problem with every starter project. In fact, I'll probably move it up in this list.

docs/testing/playwright/01-overview.md
```diff
@@ -0,0 +1,70 @@
...
+
+   ```ts title="tsconfig.json"
+   {
+     lib: [
+       ...,
+       "ESNext.Disposable"
+     ],
+     ...
+   }
+   ```
```
alicewriteswrongs, line 68: question: should the `Symbol.asyncDispose` think be possibly called out in an `info` admonition?

alicewriteswrongs (commented):
one Q

docs/testing/playwright/02-e2e-testing.md
```diff
@@ -61,6 +66,9 @@ test.describe('my-component', () => {
 });
 ```
 
+In the above snippet, the path string passed to `page.goto()` should be the _absolute_ path to the HTML file from the
+[dev server's served directory](../../config/dev-server.md#root).
```
rwaskiewicz, line 70: I'm still confused where this `my-component.e2e.html` should live - if my `root` isn't set (it has a default level of `www`), how does `my-component.e2e.html` get put in `www`?

tanner-reits:
@rwaskiewicz @alicewriteswrongs Alright, I added a step I missed (oops) to the overview/setup page and some more code snippets in the e2e testing page with an example file structure and what the test files would look like for that. I have all this running in a local repo, so we could create an demo repo in the Ionic space for a "running" example (just need someone to create the repo for me).

Depending how much is left to do in sprint work next week, I can work on getting the CLI tool I started last jedi days in a good spot so that we could automate most (or all) of the setup tasks. 

Lemme know what ya'll think

rwaskiewicz (approved):
Let's goooooo!

docs/testing/playwright/01-overview.md
```diff
@@ -0,0 +1,99 @@
...
+   the Stencil project's `www` output target config and [dev server config](../../config/dev-server.md). If no `www` output target is specified,
+   tests will not be able to run.
+
+1. Add the `copy` option to the `www` output target config:
+
+   ```ts title="stencil.config.ts"
+   {
+      type: 'www',
+      serviceWorker: null,
+      copy: copy: [{ src: '**/*.html' }, { src: '**/*.css' }]
```
alicewriteswrongs, line 80: I think there's a little typo here

```suggestion
      copy: [{ src: '**/*.html' }, { src: '**/*.css' }]
```

docs/testing/playwright/01-overview.md
```diff
@@ -0,0 +1,99 @@
...
+   ```
+
+   This will clone all HTML and CSS files to the `www` output directory so they can be served by the dev server. If you put all testing related
+   files in specific directory(s), you can update the `copy` task glob patterns to only copy those files:
+
+   ```ts title="stencil.config.ts"
+   {
+      type: 'www',
+      serviceWorker: null,
+      copy: copy: [{ src: '**/test/*.html' }, { src: '**/test/*.css' }]
```
alicewriteswrongs, line 91:
```suggestion
      copy: [{ src: '**/test/*.html' }, { src: '**/test/*.css' }]
```

alicewriteswrongs (changes requested):
noticed one thing, otherwise looks like good improvements!

alicewriteswrongs (approved):
with the changes I've now been able to stand up some simple example playwright tests (https://github.com/alicewriteswrongs/stencil-playwright-example) and it took just a few minutes to figure it out, so I think this is good to go!

PR #1383, opened 2024-03-27
chore(test): update piercing docs

christian-bromann:
__Background:__

I've discovered a limitation with Stencils deep locator and put up [a PR](https://github.com/ionic-team/stencil-site/pull/1373) to let users know about this limitation. Now, after working on https://github.com/ionic-team/stencil/pull/5481 this limitation was fixed and we can update the docs again.

__Summary:__
This patch documents that user can use multiple deep selectors in their queries.

### DIFF
--- a/docs/testing/stencil-testrunner/05-e2e-testing.md
+++ b/docs/testing/stencil-testrunner/05-e2e-testing.md
@@ -85,19 +85,11 @@ Use the "piercing" selector `>>>` to query for an object inside a component's sh
 const el = await page.find('foo-component >>> .close-button');
 ```
 
-:::caution
-Make sure to only use `>>>` once in your query. Puppeteer does not support nested deep selectors. Instead, separate the calls into two queries, e.g.:
+By default Stencil will look into all shadow roots to find the element. However if you want to restrict your query by specifying the shadow root of a particular component, you can chain multiple deep selectors within the same query, e.g.:
 
-```ts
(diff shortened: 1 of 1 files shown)


docs/testing/stencil-testrunner/05-e2e-testing.md
```diff
@@ -85,19 +85,11 @@ Use the "piercing" selector `>>>` to query for an object inside a component's sh
 const el = await page.find('foo-component >>> .close-button');
 ```
 
-:::caution
-Make sure to only use `>>>` once in your query. Puppeteer does not support nested deep selectors. Instead, separate the calls into two queries, e.g.:
+By default Stencil will look into all shadow roots to find the element. However if you want to restrict your query by specifying the shadow root of a particular component, you can chain multiple deep selectors within the same query, e.g.:
```
rwaskiewicz, line 88: Will this change only be in affect for Stencil v4.14 and up? If so, shouldn't we leave the other docs as-is?

christian-bromann: yes 🙈

christian-bromann:
@rwaskiewicz I removed the changes in older versions of the docs which remain suggesting using Puppeteer primitives.

rwaskiewicz (approved):

tanner-reits (approved):

PR #1389, opened 2024-04-01
docs(playwright): add section for migrating to/testing playwright

tanner-reits:
Adds a section to the Playwright overview for users coming from an established e2e testing setup using the Stencil Test Runner. This covers ways in which users can test out the Playwright adapter without fully migrating, or for updating file match patterns while a project is migrated to use Playwright over time.

Changes also backfilled to v4.13, v4.14, & v4.15 docs.


christian-bromann (approved):
👍

alicewriteswrongs (approved):

PR #1391, opened 2024-04-03
docs(compiler): remove docs on loading the Stencil compiler in a Web Worker or browser

christian-bromann:
It turns out this is not supported anymore since v3.3.0 of Stencil and the introduction of TypeScript v5. This patch removes these artifacts from the docs, except v3 and v2.


alicewriteswrongs (approved):
making the docs accurate! 💯

rwaskiewicz (approved):

PR #1406, opened 2024-04-16
components(api): document 'setAssetPath' API within component section

christian-bromann:
We are missing the `setAssetPath` function in the component docs. This patch adds this missing primitive.

### DIFF
--- a/docs/guides/assets.md
+++ b/docs/guides/assets.md
@@ -244,7 +244,7 @@ export declare function setAssetPath(path: string): string;
 
 Calling this API will set the asset base path for all Stencil components attached to a Stencil runtime. As a result, calling `setAssetPath` should not be done from within a component in order to prevent unwanted side effects when using a component.
 
-Make sure as component author to export this function as part of your module in order to give the consumer of your component the chance to change it, e.g. in your package entry file export the function via:
+Make sure as component author to export this function as part of your module in order to also make it accessible to the consumer of your component, e.g. in your package entry file export the function via:
 
 ```ts
 export { setAssetPath } from '@stencil/core';
--- a/docs/guides/assets.md
+++ b/docs/guides/assets.md
@@ -262,7 +262,7 @@ asset base path. This configuration depends on how your script is bundled, (or l
 
 :::note
 
-If your component library exports components compiled with [`dist-output-target`](custom-elements) and `externalRuntime` set to `true`, then `setAssetPath` has to be imported from `@stencil/core` directly.
+If your component library exports components compiled with [`dist-output-target`](/output-targets/custom-elements.md) and `externalRuntime` set to `true`, then `setAssetPath` has to be imported from `@stencil/core` directly.
 
 :::
 
--- a/docs/guides/assets.md
+++ b/docs/guides/assets.md
@@ -231,7 +231,7 @@ getAssetPath('/assets/my-image.png');
 
 ### setAssetPath
 
-`setAssetPath` is an API provided by Stencil to manually set the asset base path where assets can be found.
+`setAssetPath` is an API provided by Stencil's runtime to manually set the asset base path where assets can be found. If you are using `getAssetPath` to compose the path for your component assets, `setAssetPath` allows you or the consumer of the component to change that path.
 
 ```ts
 /**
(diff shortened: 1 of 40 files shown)


versioned_docs/version-v4.11/guides/assets.md
```diff
@@ -242,7 +242,7 @@ Calling this API will set the asset base path for all Stencil components attache
 As a result, calling `setAssetPath` should not be done from within a component in order to prevent unwanted side effects
 when using a component.
 
-If the file calling `setAssetPath` is a module, it's recommended to use `import.meta.url`.
+If the file calling `setAssetPath` is a module and your compiler is able to resolve the file location at build time, based on where the application is being deployed, you can use `import.meta.url`. In most cases however we recommend to use: `setAssetPath(\`${window.location.origin}/\`);`
```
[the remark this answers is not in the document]
christian-bromann: Yeah, seems like the change wasn't applied to that specific file. Good catch!

alicewriteswrongs (approved):

rwaskiewicz (changes requested):
I think we should add a working example here - folks have trouble with using this API, particularly when their Stencil component library is consumed elsewhere (e.g. a React library). 

E.g. https://discord.com/channels/520266681499779082/1233046047476023306/1233046047476023306

christian-bromann:
@rwaskiewicz what type of example do you have in mind? The current one I added:

```ts
import { setAssetPath } from '@stencil/core';
setAssetPath(`${window.location.origin}/`);
```

is intentionally basic as I don't want to make any assumptions on how the user is using Stencil. I could add example like this:

```html
<html>
  <head>
    <script src="https://cdn.jsdelivr.net/npm/@ionic/core/dist/ionic.js"></script>
    <script type="module">
        import { setAssetPath } from 'https://cdn.jsdelivr.net/npm/@ionic/core/dist/ionic.js';
        setAssetPath(`${window.location.origin}/`);
    </script>
  </head>
  <body>
    <ion-toggle></ion-toggle>
  </body>
</html>
```

But this would assume that the user is importing lazy components specified as `script` tag which might not be a good example for framework users.

Can you give me a hint of what type of additional example you were looking for?

rwaskiewicz:
@christian-bromann I think that's the tricky part - everyone would prefer an example for their use cases (which is one of many) 😆 

I think that an example where:
- we use it in an HTML file (with the disclaimer regarding the lazy build)
- we use it in the context of a React application

would be useful here/give folks enough information to start to figure it out for themselves in other contexts.

versioned_docs/version-v2/guides/assets.md
```diff
@@ -242,7 +242,7 @@ Calling this API will set the asset base path for all Stencil components attache
 As a result, calling `setAssetPath` should not be done from within a component in order to prevent unwanted side effects
 when using a component.
 
-If the file calling `setAssetPath` is a module, it's recommended to use `import.meta.url`.
+If the file calling `setAssetPath` is a module and your compiler is able to resolve the file location at build time, based on where the application is being deployed, you can use `import.meta.url`. In most cases however we recommend to use: `setAssetPath('${window.location.origin}/');`
```
rwaskiewicz, line 245: Can we roll this back, since we're not maintaining v2 docs anymore?

versioned_docs/version-v3/guides/assets.md
```diff
@@ -209,7 +211,8 @@ declare function getAssetPath(path: string): string;
 ```
 
 The code sample below demonstrates the return value of `getAssetPath` for different `path` arguments, when an asset base path of `/build/` has been set.
-```ts
+import { getAssetPath } from '@stencil/core';
+
```
rwaskiewicz, line 215: I think this breaks the code styling - we lose the beginning of the code block here (it looks like this happens a few other places in this PR as well) - My guess is this is what's causing the MD link checks to fail as well. Since the code blocks are now 'offset', in that the ends of code blocks are now the starts of code blocks, pre-existing headers are assumed to be a part of a code block

docs/guides/assets.md
```diff
@@ -242,9 +246,29 @@ Calling this API will set the asset base path for all Stencil components attache
...
+asset base path. This configuration depends on how your script is bundled, (or lack of bundling), and where your assets can be loaded from.
+
+For example, in case you import a component directly via script tag, this would look like:
+
+```html
+<html>
+  <head>
+    <script src="https://cdn.jsdelivr.net/npm/@ionic/core/dist/ionic.js"></script>
+    <script type="module">
+      import { setAssetPath } from 'https://cdn.jsdelivr.net/npm/@ionic/core/dist/ionic.js';
```
rwaskiewicz, line 266: If I copy/paste this into a local file, I get "import not found: setAssetPath" - is this the bug you were referring to in Slack this morning/last night?
Screenshot 2024-05-01 at 8 54 06 AM

docs/guides/assets.md
```diff
@@ -242,9 +246,29 @@ Calling this API will set the asset base path for all Stencil components attache
 As a result, calling `setAssetPath` should not be done from within a component in order to prevent unwanted side effects
 when using a component.
 
-If the file calling `setAssetPath` is a module, it's recommended to use `import.meta.url`.
+If the file calling `setAssetPath` is a module and your compiler is able to resolve the file location at build time, based on where the application is being deployed, you can use `import.meta.url`. In most cases however we recommend to use:
```
rwaskiewicz, line 249: I'm a little confused by this part:

> your compiler is able to resolve the file location at build time

I thought this was a runtime construct/action, which we seem to imply this above this line with:

> Calling this API will set the asset base path for all Stencil components attached to a Stencil runtime.
As a result, calling `setAssetPath` should not be done from within a component in order to prevent unwanted side effects
when using a component.

Do I have that right? Or is there some build time actions we do that I don't recall (equally possible 😆)

rwaskiewicz (changes requested):

versioned_docs/version-v3/guides/assets.md
christian-bromann: uff .. I think this is a copy&paste issue from my side.

docs/guides/assets.md
christian-bromann: Correct, I will try to address this in the docs until we have a fix for this.

docs/guides/assets.md
christian-bromann: I am referring here to e.g. an external compiler that a user is using to create application bundle. In hindsight this seems indeed very ambiguous, I will try to rephrase this.

rwaskiewicz (approved):
LGTM! Thanks!

PR #1407, opened 2024-04-18
docs(vitest): add Vitest option to testing section

christian-bromann:
This patch adds another option for testing Stencil components using Vitest. While this option has its limitation (no auto-waits and no mocking) it still is an option that is useable and worth to document.


docs/testing/03-vitest.md
```diff
@@ -0,0 +1,183 @@
...
+    },
+    plugins: [stencil()]
+})
+```
+
+This configuration enables tests to run in a headless Chrome browser.
+
+## Writing Tests
+
+Once you've setup Vitest you can start write your first test. In order to render a Stencil component into the browser, all you need to do is importing the component and initiating an instance of the component on the page:
```
alicewriteswrongs, line 51:
```suggestion
Once you've setup Vitest you can start write your first test. In order to render a Stencil component into the browser, all you need to do is import the component and initiate an instance of the component on the page:
```

docs/testing/03-vitest.md
```diff
@@ -0,0 +1,183 @@
...
+
+```ts
+// src/components/my-component/my-component.test.ts
```
alicewriteswrongs, line 54: we could use the little `title` option like so if we want here:

````suggestion
```ts title="src/components/my-component/my-component.test.ts"
````

that goes for the other filenames in here as well

File: docs/testing/03-vitest.md [hunk 3]
```diff
@@ -0,0 +1,183 @@
...
+        "test": "vitest --run"
+    },
+}
+```
+
+Execute the tests using:
+
+```sh
+npm test
+```
```
alicewriteswrongs, line 85: we can also put the `npm2yarn` directive on here

````suggestion
```bash npm2yarn
npm test
```
````

File: docs/testing/03-vitest.md [hunk 4]
```diff
@@ -0,0 +1,183 @@
...
+position: 5
+---
+
+# Overview
+
+[Vitest](https://vitest.dev/) is a popular and modern test framework for unit testing. You can use Vitest to test Stencil components in the browser using its [browser mode feature](https://vitest.dev/guide/browser.html).
+
+:::note
+Vitest browser mode is an experimental feature and in early development. As such, it may not yet be fully optimized, and there may be some bugs or issues that have not yet been ironed out.
+:::
```
alicewriteswrongs, line 12: should this maybe be a `caution`? Maybe that's too alarmist haha

alicewriteswrongs (changes requested):
a good addition I think! Just noticed a few things we might want to tweak

christian-bromann:
@alicewriteswrongs thanks for your feedback, addressed all comments.

alicewriteswrongs (approved):

tanner-reits (approved):

PR #1410, opened 2024-04-22
docs(playwright): add note for Stencil config naming limitation

tanner-reits:
Adds a note admonition to the Playwright adapter docs for the naming limitation around the Stencil config. This note already exists in the adapter readme, so just porting it over here as well.


rwaskiewicz (approved):

christian-bromann (approved):
👍

PR #1419, opened 2024-05-01
docs(testing): add guidance to better test when using async global scripts

christian-bromann:
This patch is a result of a discussion in https://github.com/ionic-team/stencil/issues/5526 where a community member suggested this approach to wait for the components to be hydrated. I've attempted to embed this into Stencil itself but due to the fact that the hydration flag is based on a Stencil configuration that we don't have access too, this wouldn't be a general approach.

Hence, I added this piece of documentation which hopefully can guide users better when using async global scripts.

### DIFF
--- a/docs/testing/stencil-testrunner/05-e2e-testing.md
+++ b/docs/testing/stencil-testrunner/05-e2e-testing.md
@@ -225,6 +225,18 @@ const el = await page.find('.desktop');
 expect(el).not.toBeDefined();
 ```
 
+### Asynchronous Global Script
+
+If you are using an asynchronous [global script](/config/01-overview.md#globalscript), it may happen that your Stencil components haven't been hydrated at the time you are trying to interact with them. In this case it is recommend to wait for the hydration flag to be set on the component, e.g.:
+
+```ts
+const page = await newE2EPage();
+await page.setContent(`<my-cmp></my-cmp>`);
+await page.waitForSelector('.hydrated')
+```
+
+__Note:__ the hydrate selector may be different depending on whether you have a custom [`hydratedFlag`](/config/01-overview.md#hydratedflag) options set.
+
 ## Caveat about e2e tests automation on CD/CI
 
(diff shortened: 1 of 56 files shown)


docs/testing/stencil-testrunner/05-e2e-testing.md
```diff
@@ -225,6 +225,18 @@ const el = await page.find('.desktop');
...
+
+If you are using an asynchronous [global script](/config/01-overview.md#globalscript), it may happen that your Stencil components haven't been hydrated at the time you are trying to interact with them. In this case it is recommend to wait for the hydration flag to be set on the component, e.g.:
+
+```ts
+const page = await newE2EPage();
+await page.setContent(`<my-cmp></my-cmp>`);
+await page.waitForSelector('.hydrated')
+```
+
+__Note:__ the hydrate selector may be different depending on whether you have a custom [`hydratedFlag`](/config/01-overview.md#hydratedflag) options set.
```
tanner-reits, line 238: Can turn this into an admonition for better vis:
```suggestion
:::note
The hydrate selector may be different depending on whether you have a custom [`hydratedFlag`](/config/01-overview.md#hydratedflag) options set.
:::
```

tanner-reits (commented):
One suggestion. Also, do we need to backfill this all the way to v4.0? Can't remember when we fixed/supported async scripts (or if that was always a thing and I'm remembering incorrectly 😅)

docs/testing/stencil-testrunner/05-e2e-testing.md
christian-bromann: 👍

christian-bromann:
> Also, do we need to backfill this all the way to v4.0?

Good catch! Remove changes and only introduce them with [v4.12.0](https://github.com/ionic-team/stencil/releases/tag/v4.12.2) which is where we fixed this behavior.

rwaskiewicz (approved):

tanner-reits (approved):

PR #1444, opened 2024-06-13
docs(config): document 'env' configuration

christian-bromann:
fixes [#1611](https://github.com/ionic-team/stencil/issues/1611)

### DIFF
--- a/docs/config/01-overview.md
+++ b/docs/config/01-overview.md
@@ -125,7 +125,7 @@ export const config: Config = {
 }
 ```
 
-Now when you build your components which this environment variable set, you can import it in your component as following:
+Now when you build your components with this environment variable set, you can import it in your component as follows:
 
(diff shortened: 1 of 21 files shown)


docs/config/01-overview.md
```diff
@@ -108,6 +108,39 @@ enableCache: true
...
+
+export const config: Config = {
+  ...,
+  env: {
+    API_BASE_URL: process.env.API_BASE_URL
+  }
+}
+```
+
+Now when you build your components which this environment variable set, you can import it in your component as following:
```
tanner-reits, line 128:
```suggestion
Now when you build your components with this environment variable set, you can import it in your component as follows:
```

tanner-reits (approved):
Little grammatical fix, but otherwise good to go!

PR #1452, opened 2024-06-24
docs(compiler): document transformTagName feature

christian-bromann:
This feature was introduced in https://github.com/ionic-team/stencil/pull/2343 and never documented.


christian-bromann:
Nevermind, this feature doesn't seem to work properly across lazy loading and custom components. Let's not introduce it.

PR #1461, opened 2024-07-16
chore(renovate): lower update frequency

tanner-reits:
Lowers the frequency of renovate updates to once a month


christian-bromann (approved):
There is one linting issue which we should resolve before merge but the change LGTM 👍

Issue #1465, opened 2024-07-29
`getElement` doc type mismatch


nwhittaker:
The `getElement` type published to the doc site does not match the `getElement` type in `stencil-public-runtime.d.ts`.

| Doc site | `stencil-public-runtime.d.ts` |
| - | - |
| `(ref: getElement) => string` | `(ref: any) => HTMLStencilElement` |

Expecting the doc site to match the runtime type.

PR #1464, opened 2024-07-18
docs(copy-task): Add docs to copy task options

christian-bromann:
Adding docs for all parameters the copy task takes.

PR #1466, opened 2024-07-29
Fix `getElement` type in doc

nwhittaker:
Closes https://github.com/ionic-team/stencil-site/issues/1465.

### DIFF
--- a/docs/components/api.md
+++ b/docs/components/api.md
@@ -184,7 +184,7 @@ The following primitives can be imported from the `@stencil/core` package and us
 
 - **getElement()**: Retrieve a Stencil element for a given reference.
 
-  __Type:__ `(ref: getElement) => string`<br />
+  __Type:__ `(ref: any) => HTMLStencilElement`<br />
   __Example:__
   ```ts
   import { getElement } from '@stencil/core'
--- a/docs/components/api.md
+++ b/docs/components/api.md
@@ -123,7 +123,7 @@ The following primitives can be imported from the `@stencil/core` package and us
 
 - **forceUpdate()**: Schedules a new render of the given instance or element even if no state changed. Notice `forceUpdate()` is not synchronous and might perform the DOM render in the next frame.
 
-  __Type:__ `(ref: HTMLElement) => void`<br />
+  __Type:__ `(ref: any) => void`<br />
   __Example:__
   ```ts
   import { forceUpdate } from '@stencil/core'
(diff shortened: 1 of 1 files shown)


docs/components/api.md
```diff
@@ -184,7 +184,7 @@ The following primitives can be imported from the `@stencil/core` package and us
 
 - **getElement()**: Retrieve a Stencil element for a given reference.
 
-  __Type:__ `(ref: getElement) => string`<br />
+  __Type:__ `(ref: any) => HTMLStencilElement`<br />
```
christian-bromann, line 187: Shouldn't it be rather
```suggestion
  __Type:__ `(ref: HTMLElement) => HTMLStencilElement`<br />
```

christian-bromann:
Thanks for raising the PR 🙏  one comment on the argument type.

docs/components/api.md
nwhittaker: I went with `any` to match the type in [`stencil-public-runtime.ts`][1]. That being said, I agree it'd be great to further narrow the type. I see `(ref: HTMLElement) => HTMLStencilElement` matches the example, but reading through the other docs and examples on the api page, it seems like a `ref` can refer to either an `HTMLElement` or a `ComponentInterface`.

For example, `forceUpdate` and `getMode` are typed on the doc page as `(ref: HTMLElement) => void`, but their example code shows passing in a component instance. Similarly, `setMode` is typed the same but its example show passing in an arrow function.

@christian-bromann, I'm happy to fix up the types and examples, but I need to know how to type `ref`. Is `ComponentInterface | HTMLElement` sufficient? Would you anticipate any problems with reflecting this type back to the https://github.com/ionic-team/stencil repo?

[1]: https://github.com/ionic-team/stencil/blob/c3d4e8b170b405ef420236f12d7b19e21e541a81/src/declarations/stencil-public-runtime.ts#L337

christian-bromann: I think the reason why people put `any` in there in the first place was because the actual real type is not suppose to be exposed for public consumption. Hence, let's keep it close to how it is currently in the Stencil public documentation and disregard my comment here.

christian-bromann (commented):
One last ask: mind copying this change to all versioned docs (except v2 and v3 docs which are deprecated Stencil versions), e.g. like in #1464

nwhittaker:
>One last ask: mind copying this change to all versioned docs (except v2 and v3 docs which are deprecated Stencil versions)

@christian-bromann, I can do that. I added a commit to correct the `forceUpdate`, `setMode`, and `getMode` types. Can you please verify those before I propagate the changes to other doc versions?

christian-bromann:
@nwhittaker thanks, this all looks good to me.

christian-bromann (approved):

PR #1467, opened 2024-07-31
docs(config): add watchIgnoredRegex

tanner-reits:
Adds documentation for the `watchIgnoredRegex` Stencil config option.

I do not intend to backfill these changes to previous versions since v4.20 will be the first version to support tsconfig watchOptions

### DIFF
--- a/docs/config/01-overview.md
+++ b/docs/config/01-overview.md
@@ -523,9 +523,9 @@ Passes custom configuration down to rollup itself. The following options can be 
 
 ## watchIgnoredRegex
 
-_default: `[]`_
+*default: `[]`*
 
-_type: `RegExp | RegExp[]`_
+*type: `RegExp | RegExp[]`*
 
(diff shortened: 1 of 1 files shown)


docs/config/01-overview.md
```diff
@@ -519,4 +519,29 @@ Passes custom configuration down to rollup itself. The following options can be
...
 
-*default: `{}`*
\ No newline at end of file
+*default: `{}`*
+
+## watchIgnoredRegex
+
+_default: `[]`_
+
+_type: `RegExp | RegExp[]`_
```
christian-bromann, line 528: Just to be aligned with previous formatting even though it will turn out to be the same
```suggestion
*default: `[]`*

*type: `RegExp | RegExp[]`*
```

christian-bromann (approved):
One comment but overall LGTM 👍

PR #1482, opened 2024-09-17
Update methods.md

msqaddura:
it seems what we want to say is that;  it is callable from outside, not directly from the element (component).
though they can be called directly but the following sentence "i.e. they are intended to be callable from the outside!"
states otherwise

### DIFF
--- a/versioned_docs/version-v4.21/components/methods.md
+++ b/versioned_docs/version-v4.21/components/methods.md
@@ -7,7 +7,7 @@ slug: /methods
 
 # Method Decorator
 
-The `@Method()` decorator is used to expose methods on the public API. Functions decorated with the `@Method()` decorator can not be called directly from the element, i.e. they are intended to be callable from the outside!
+The `@Method()` decorator is used to expose methods on the public API. Functions decorated with the `@Method()` decorator should not be called directly from the element, i.e. they are intended to be callable from the outside!
 
 :::note
 Developers should try to rely on publicly exposed methods as little as possible, and instead default to using properties and events as much as possible. As an app scales, we've found it's easier to manage and pass data through @Prop rather than public methods.
--- a/versioned_docs/version-v4.21/components/methods.md
+++ b/versioned_docs/version-v4.21/components/methods.md
@@ -7,7 +7,7 @@ slug: /methods
 
 # Method Decorator
 
-The `@Method()` decorator is used to expose methods on the public API. Functions decorated with the `@Method()` decorator can be called directly from the element, i.e. they are intended to be callable from the outside!
+The `@Method()` decorator is used to expose methods on the public API. Functions decorated with the `@Method()` decorator can not be called directly from the element, i.e. they are intended to be callable from the outside!
 
 :::note
 Developers should try to rely on publicly exposed methods as little as possible, and instead default to using properties and events as much as possible. As an app scales, we've found it's easier to manage and pass data through @Prop rather than public methods.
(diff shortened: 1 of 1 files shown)


christian-bromann:
@msqaddura mhm 🤔 I am not sure if this makes sense, do you have a concrete example?

msqaddura:
maybe first let me explain myself. to be sure i understand it correctly

1. so the point of `@Method` decorator is to give access to components, from another component. and that is when we use `document.querySelector`. is that right?

2. a component doesnt need `@Method` to access its own functions, we can use `private func...` or `public func...`

3. the statement below might be misleading given the example is not related to the statement first part says it can be called from within, the i.e given states that it is called from outside

> decorator can be called directly from the element, i.e. they are intended to be callable from the outside!


/* @jsx h */
import { h, Component, Host, State, Method } from '@stencil/core';

@Component({
  tag: 'with-method',
  shadow: true,
})
export class WithMethod {
  @State() text = '';
  @State() counter = 0;

  @Method() // Method is needed to expose the function
  async setTextFromOutside(text: string) {
    this.text = text;
  }

  // same functionality can be achieved without @Method
  private setTextWithin(text: string) {
    this.text = text;
  }
  onClick() {
    this.counter += 1;
    this.setTextWithin('Called Within!'); // this call can be private
    this.text = 'Called Within!'; //we can access name from here, no need for setText
  }

  render() {
    return (
      <Host>
        <button onClick={() => this.onClick()}>Click me</button>
      </Host>
    );
  }
}

@Component({
  tag: 'method-caller',
  shadow: true,
})
export class MethodCaller {
  onClick() {
    const target = document.querySelector('with-method')!;

    target.setTextFromOutside('Called Outside!');
    // target.setTextWithin("Error"); //this is not going to work as it is not exposed via @Method()
  }
  render() {
    return (
      <Host>
        <button onClick={() => this.onClick()}>Click me</button>
      </Host>
    );
  }
}


Overall, maybe it is not that important but would be good to explain the intent of the `@Method` that it is meant to be called from outside the component

msqaddura:
@christian-bromann 
maybe as well let me ask about the use-case for `@Method`
if my understanding is correct, it is meant to be used when we want to do something inside a component using javascript (by using `document.querySelector`).

however, i might have limited it to updating properties/state from outside. and there might be another use-case for it.

so maybe it is more about me misunderstanding the motive behind `@Method` :D

christian-bromann:
Functions decorated with `@Method` are suppose to be public functions that you can call via e.g.:
```ts
(async () => {
  await customElements.whenDefined('todo-list');
  const todoListElement = document.querySelector('todo-list');
  await todoListElement.showPrompt();
})();
```
Now you change says:
> Functions decorated with the `@Method()` decorator should not be called directly from the element

Why should it not be called from the element?

christian-bromann:
Any feedback @msqaddura ?

msqaddura:
I will give it one more shot, if not then we can close the ticket 😃

imagine a new developer wants to use Stencil, and he comes to read this
> Functions decorated with the @Method() decorator can be called directly from the element, **i.e.** they are intended to be callable from the outside!

now lets replace  **i.e** with **however**.
> Functions decorated with the @Method() decorator can be called directly from the element, **however** they are intended to be callable from the outside!

How does it feel to replace **i.e** with **however**? 
-------------------------------------------
I think overall, what I am trying to say is that the wording might easily trick people. I have seen many devs with confused ideas regarding the purpose of `@Method` 
After reading the docs, it came to my mind that the wording could be the problem.

If we agree on the rewording, then we can expand the topic 😃 If not, then we can close the ticket. as I dont want to consume your time

Other than that, thank you for stencil! <3 and keep the good work up! personally I am in love with it

PR #1494, opened 2024-11-05
react: add docs for excludeServerSideRenderingFor option

christian-bromann:
Companion PR for https://github.com/ionic-team/stencil-ds-output-targets/pull/540

I also removed the note saying that SSR support only works for Next.js App Router which is not true. The type of router shouldn't matter as long as the user uses Next.js v13 and above.

### DIFF
--- a/docs/framework-integration/react.md
+++ b/docs/framework-integration/react.md
@@ -508,7 +508,7 @@ export const config: Config = {
 
 **Type: `string[]`**
 
-Allows user to exclude a list of components to be server side rendered by Next.js or other React
+Allows users to exclude a list of components from server side rendering by Next.js or other React
 frameworks. This may be useful if you would like to generally ignore some components from being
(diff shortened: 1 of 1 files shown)


docs/framework-integration/react.md
```diff
@@ -502,9 +502,16 @@ export const config: Config = {
...
-:::note
-Next.js support is only available for applications that use the [Next.js App Router](https://nextjs.org/docs/app).
-:::
+### excludeServerSideRenderingFor
+
+**Optional**
+
+**Type: `string[]`**
+
+Allows user to exclude a list of components to be server side rendered by Next.js or other React
```
tanner-reits, line 511:
```suggestion
Allows users to exclude a list of components from server side rendering by Next.js or other React
```

tanner-reits (approved):
One tweak, non blocking

PR #1497, opened 2024-12-02
undo dist.md 'Distribution Options' docs

johnjenkins:
Docs regarding 'Distribution Options' in Stencil were (I believe) wrongly removed.

1) they're useful
2) they're mentioned in remaining text - "loading the components directly through a script tag as mentioned below"


christian-bromann (approved):
The more docs the merrier 😉

PR #1506, opened 2025-02-07
Update docs for new puppeteer version

s-gurevich:
1. Adapted the documentation referencing to Stencil config file, because a property was changed in the recent updates
2. Added new version of puppeteer to the support site

Background:
Chrome changed how they handle headless browser distribution: https://developer.chrome.com/blog/chrome-headless-shell
Puppeteer adapted their config: https://pptr.dev/guides/headless-modes
Stencil upgraded the puppeteer dependency and adapted their config.

`headlessBrowser: 'new'` is no longer accepted.


christian-bromann (approved):
Thank you so much 👍

PR #1523, opened 2025-05-10
Update server-side-rendering.md

kevinblilly:
The export configuration for the hydrate module are inaccurate. There is no `index.cjs.js` file being produced by the `dist-hydrate-script` output target.


christian-bromann (commented):
Can we also apply the change to the latest docs `docs/guides/server-side-rendering.md`?

christian-bromann (approved):
👍

PR #1527, opened 2025-06-02
docs(stencil-docs): adding clarifying info boxes for the  command

ShaneK:
While investigating solutions to [core/#5883](https://github.com/stenciljs/core/issues/5883), I found that the current behavior of the `stencil docs` command seems intentional and that generating CSS docs would require a build, which the `stencil docs` command does not due.

After talking to Christian about this, we decided that, for now, the approach to resolve this issue would be updating the docs to clarify this behavior so prevent confusion in the future.

In the future, we will rework the Stencil CLI so that things like this are less confusing and more intuitive.


christian-bromann (approved):
LGTM 👍 

Great work!

PR #1531, opened 2025-06-05
Update vue.md

jvlobo:
'externalEvent' does not exist in type 'ComponentModelConfig'


christian-bromann:
@jvlobo let's update it for the all versions upwards v4.31

PR #1535, opened 2025-06-18
Update 04-mocking.md

smallbonelu:
fix typo issue


smallbonelu:
> Good catch 👌
> 
> Can we also make the change for the next docs in `/docs/testing/stencil-testrunner/04-mocking.md`?

I've created another PR for this file.

PR #1539, opened 2025-07-06
Update react.md

AlbertGabdullin:
There is missed `import { Config } from "@stencil/core";`


christian-bromann (changes requested):
Can we update this in all versioned files including `/docs/framework-integration/react.md`?

PR #1569, opened 2025-12-17
docs: add Extends & Mixins components architecture guide

paulvisciano:
Adds documentation for the extends and Mixin() features introduced in Stencil v4.37.0, covering component architecture patterns using controllers.

- New documentation page: docs/components/extends.md (also added to versioned_docs/version-v4.39/components/)

Guide covers:

- Using extends and Mixin() for component architecture
- Inheritance vs composition patterns
- When to use each approach
- Scaling considerations for different library sizes
- Code examples for both patterns
- ReactiveControllerHost implementation reference


gnbm (approved):
LGTM 🚀

paulvisciano:
> I think this should probably go under `guides` as it's less about Stencil's API. No real hard feelings about it though.
> 
> More important are comments received (where I tend to agree) where the new docs feel quite hard to read and understand. They're quite obviously LLM generated - can you make it more readable and approachable?
> 
> You want devs to 'grok' new features and the power they offer in the simplest of terms; the sheer length of the doc is off putting.

Great feedback, thank you. Moved it to guides & made it a lot shorter and easier to follow, take another look when you get a chance. 

Also I am getting a Vercel Auth required for deploy, even though I authorized it. Any pointers on resolving that?

image

